Issue #19, opened 2014-04-22
Add -c option for alternative config file


omangin:

73:
Hi,
I have some code on this one but it seems that I can't get my head around the FakeFilesystem for testing. 
The statement:
```python
self.fs['os'].path.exists(self.fs['os'].path.expanduser('~/.pubsrc'))
```
always returns `false`. 
The users home should be initialized https://github.com/pubs/pubs/blob/master/tests/fake_env.py#L116 right? 
But the config file does not seem to reside inside the fake file system. 
To recreate just add the line to a test with an `pubs init`.

What am I missing here?
--73

humm:
Yes, FakeFilesystem is kinda tricky. 

If you modified the code in ways that add calls to the filesystem in a new module, you must add it to:
https://github.com/pubs/pubs/blob/master/tests/test_usecase.py#L59
and similar lines in  https://github.com/pubs/pubs/blob/master/tests/test_databroker.py

Frankly, it needs to be more robust to changes, I'll add that to the todo list. 

If that does not solve your problem, I need to see your code to help you more.

73:
Well, here it is: https://github.com/73/pubs/commit/2e0fca5430c721a22c42e151f41c27fca890522d

The configs are already added. `open()` should be called inside the `configobj` that might be the issue, but I'm really not sure.  

Thanks,
--73

humm:
It appear you have brought old code along with your changes when rebasing against the last commit of the `feat/new_config` branch. Your problems should be solved by a cleaner patch. The `feat/new_config` branch is almost finished, but wide-ranging color changes, and a merge with the new master code will still make it a highly moving target for a day or two.

73:
Thx for the feedback, I'll clean up.

humm:
Ok, I pushed my color changes and merged with master. The code should be relatively stable now. It will be pushed into master soon.

73:
Nice. I ran into a problem which I can't decide for myself. 
Currently I can't call `config.get_confpath()` to determine the correct order of precedence for the config file (which I guess would be switch > env > default), because the function relies on `uis.get_ui()` https://github.com/pubs/pubs/blob/feat/new_config/pubs/config/conf.py#L30. But I need a config file to create it https://github.com/pubs/pubs/blob/feat/new_config/pubs/uis.py#L31-L39. Therefore there is an dependency cycle. From the top of my head I see to options to resolve this:
[...]

PR #36, opened 2015-12-04
feature #19

73:
not that I need this feature, but i like this project. :)


73:
I missed a test...

humm:
Hi, nice work. Glad you like the project. 

Just a heads-up, the new config code is progressing, it's here: [feat/new_config](/pubs/pubs/tree/feat/new_config). I need to polish it a bit more before I merge it into master, but it will happen soon.

73:
I will rebase against feat/new_config. I think that makes the most sense.

humm:
Probably best. A commit just came your way.

PR #37, opened 2015-12-07
feat/new_config: better, more robust, more flexible configuration

humm:
@omangin 
Can you check that it works for you before I push everything in master ?


omangin:
That works with me, with 0bfe921.

The test not passing are another issue: one test behaves unpredictably by randomly failing due to ordering of the results in the dictionary.

humm:
Ok, the work of @73 has been merged in the branch. Before merging to master, I want to make the alias plugin work again, because it is in the readme. Some fixes have been made in the last commit, but more work remains.

omangin:
Just fixed the issues with the alias plugin.

As you predicted, @humm, I also disabled the config update code: the config would continually be rewritten, overriding the plugin's configuration. The check is now only run with the `--update` flag.  I would be in favor of removing the update code which has little benefit and will most probably be only used once or twice. Maybe a documentation of the new config (at least a pointer to the self-documented config) is a better solution.

The only issue is that I disabled the update by disabling the check, there is no check any more. Maybe we have to fix it before merging. It is good to have the code complaining about wrong configuration, at least as a warning.

humm:
Ok, I fixed the update overriding the plugins section. Upgrade will also only happen when the version is bumped, and when we do release a new version, we can decide if an upgrade is needed or not. I removed the --upgrade flag, that is not needed anymore. I am merging everything in master.

humm:
About the update mechanism: I do think we can improve it to the point it is efficient and unobtrusive, thus reducing the number of flag/command/and obsolete configurations we have to handle.

PR #38, opened 2015-12-08
implements #19

73:
I guess this is ready to be merged now...

### DIFF
--- a/pubs/config/conf.py
+++ b/pubs/config/conf.py
@@ -2,6 +2,7 @@ import os
 import platform
 import shutil
 
+
 import configobj
 import validate
 
@@ -10,12 +11,14 @@ from .spec import configspec
 
 DFT_CONFIG_PATH = os.path.expanduser('~/.pubsrc')
 
+
 def load_default_conf():
     """Load the default configuration"""
     default_conf = configobj.ConfigObj(configspec=configspec)
     validator = validate.Validator()
     default_conf.validate(validator, copy=True)
     return default_conf
+
 
 def get_confpath(verify=True):
     """Return the configuration filepath
@@ -32,11 +35,13 @@ def get_confpath(verify=True):
             ui.exit(error_code=1)
     return confpath
 
+
 def check_conf(conf):
     """Type check a configuration"""
     validator = validate.Validator()
     results   = conf.validate(validator, copy=True)
     assert results == True, '{}'.format(results) # TODO: precise error dialog when parsing error
+
 
 def load_conf(check=True, path=None):
     """Load the configuration"""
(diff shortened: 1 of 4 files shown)


humm:
Hum... seems this is not compliant with python 3---see the Travis CI report. As we have users using both versions of python, this is a really a requirement for any pull request.

73:
I see. Shouldn't be that hard to make it runnable with 2.7, 3.3-5 ... Do you have a tip on how to check this locally? I may just run PyCharm code inspection, but I'm not sure I will always have an IDE at hand.

humm:
You just need to have an another version of python you can switch to for the tests. I personally use [pyenv](http:/github.com/yyuu/pyenv), but there are other ways, such as using `virtualenv`.

73:
Hey, 
so I used pyenv (virtualenv only provides my distros python versions) and went ahead and installed python 3.5.0 (which is exactly the same version that travis uses).  And ran the `nosetests`  like travis did. 
```
[~/repo/pubs][⎇ ][●][17:41]
[$] python --version
Python 3.5.0

[~/repo/pubs][⎇ ][●][17:41]
[$] nosetests
........................................................................................
----------------------------------------------------------------------
Ran 88 tests in 1.689s

OK
```
So there are some options:
1: I could fix the error that travis gives "blindly" and hope it works (because it works on my box either way). 
2: You could check out and run the tests (maybe you can confirm travis).
3: You have another idea... (which I'm sure will be better then mine)....

**TL;DR** *unable to reproduce*, what to do now?

humm:
I see. I am going to look if I can reproduce it on my machine tomorrow. I'll keep you updated.

omangin:
OK. It seems that the `version` argument for `ArgumentParser` was dropped. It seems that it was originally for compatibility with *optparse*.

You have to replace it with an explicit call to  `add_argument`.

See last line from <https://docs.python.org/2.7/library/argparse.html#upgrading-optparse-code>.

73:
@omangin thx for having a look. I found that already. But if I am not able to reproduce the error travis is throwing, I have no reliable way to check whether or not it is the only error.

humm:
Ok, I pulled your commit, rebased on fixes introduced by 0bfe921a, and added commit to make it python 3 compliant. You can see your work in the feat/new_config branch, soon in master. Thanks for all your work. I'll probably build on your changes to catch 'init' with the conf_parser.

humm:
@73 
I want to add your name to the list of contributors. Should I add it as 73 or should I use your real name ?

73:
You can use my real one: Arnold Sykosch

PR #52, opened 2016-01-13
Better exceptions handling at the ui level

omangin:
This branch implements a simple exception handling at the ui level. One goal is that no error stack reaches the user any more. This code enables to remove generic catching of exceptions (typically redundant print of the same error message and exit in each command). Now exceptions are catched in the execute function and handled by the ui. The ui decides (depending on whether a *debug* mode is active) if the exception should be re-raised or transformed into an error message followed by exit with non-zero error code.

The whole system relies in setting meaningful error messages (when possible) for exceptions raised at the low level. The exception is meant to carry the message until some code can deal with it, which will ultimately be the case of the ui.

From 8fafdc0:
>The default behavior for commands is now to only catch exceptions that
must be handled specifically. This includes outputting a context
dependent message, cleaning up, etc. All other exceptions will be
handled by the ui.

The way the message is extracted from exceptions is by using the `__str__` method which can be overridden. For example the citekeys related exception retain the guilty citekey as an attribute in case a top level loop would want to catch the exception and know which citekey failed. However this is not used yet. The fact is that it may be better to use the `message` attribute instead but I have no strong opinion on the matter. Please tell if you have one :)

Finally I plan on implementing a mapping from standard exceptions to error codes, which may be interesting for scripting purposes (for example a dedicated error code for citekey related errors would be informational). Do you have any opinion on that matter too?

Typical error codes would be like the following.
```python
DEFAULT = 1
WRONG_CITEKEY = 3
FILE_NOT_FOUND = 4
CANNOT_OPEN_FILE = 5
PARSING_FAILED = 6
NO_RESULT = 7
WRONG_QUERY = 2
UNABLE_TO_RUN_COMMAND = 127
``` 

Still have to handle `SIGINT` correctly too.

### DIFF
--- a/pubs/commands/add_cmd.py
+++ b/pubs/commands/add_cmd.py
@@ -104,11 +104,7 @@ def command(conf, args):
         ui.error('citekey already exist {}.'.format(citekey))
         ui.exit(1)
 
-    try:
-        p = paper.Paper.from_bibentry(bibentry, citekey=citekey)
-    except Exception as e:
-        ui.error(e.args[0])
-        ui.exit(1)
+    p = paper.Paper.from_bibentry(bibentry, citekey=citekey)
 
     # tags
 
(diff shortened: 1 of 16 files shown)


humm:
Looks good to me. Nice work.

I think we could take this opportunity to properly add docstrings about the semantics of the exceptions classes we are creating (however obvious they may seem). 

The message attribute is deprecated since python 2.6 (see this discussion for instance http://stackoverflow.com/questions/1272138/baseexception-message-deprecated-in-python-2-6). I think it is better to store the message in the exception argument rather than override the `__str__()` function. This is more flexible, and would allow to pass more complex data than string for formatting, which could be also helpful during debug. 

No opinion on the error codes. Not really familiar with the subject, but that seems a good idea. 

PS: Seems pyfakefs is having trouble... (https://github.com/jmcgeheeiv/pyfakefs/issues/45)

PR #53, opened 2016-01-17
Fix expectedFailure tests

humm:


humm:
@omangin: could you verify that this fixes expectedFailure tests on your end? Expected failure tests are really good to functionally describe bugs before fixing them.

omangin:
It works. I merge.

PR #54, opened 2016-01-23
Feat/cache

humm:
@omangin:
These commits are the single most involved I had to write in a long time. It required diving in pyfakefs, configobj and python 2 and 3 source code. And doing a PR on pyfakefs to implement nanosecond mtime on the fake filesystem.

I only tested it---outside of the fake filesystem---on my system, using HFS+, that only implements integers stat times (I take solace in the fact that HFS+ only represents dates up to 2040, so Apple *will* be forced to move on to a more modern filesystem in the next 25 years).

Could you do that on your box? Do make a backup.

BTW, pubs is fast again!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
   - "3.5"
 # command to install dependencies
 install:
-    - "pip install pyfakefs"
+    - "pip install -r tests/requirements.txt"
     - "python setup.py install"
 # command to run tests
 script: python -m unittest discover
--- a/pubs/apis.py
+++ b/pubs/apis.py
@@ -17,8 +17,7 @@ def isbn2bibtex(isbn):
 
     url = 'http://www.ottobib.com/isbn/{}/bibtex'.format(isbn)
     r = requests.get(url)
-    soup = BeautifulSoup(r.text)
+    soup = BeautifulSoup(r.text, "html.parser")
     citation = soup.find("textarea").text
 
     return citation
-
--- a/pubs/commands/add_cmd.py
+++ b/pubs/commands/add_cmd.py
@@ -141,3 +141,5 @@ def command(conf, args):
                 ui.message('{} was moved to the pubs repository.'.format(docfile))
             else:
                 ui.message('{} was copied to the pubs repository.'.format(docfile))
+
+    rp.close()
--- a/pubs/commands/doc_cmd.py
+++ b/pubs/commands/doc_cmd.py
@@ -148,3 +148,5 @@ def command(conf, args):
         except OSError:
             ui.error("Command does not exist: %s." % with_command)
             ui.exit(127)
+
+    rp.close()
--- a/pubs/commands/edit_cmd.py
+++ b/pubs/commands/edit_cmd.py
@@ -67,3 +67,5 @@ def command(conf, args):
                 break
             # else edit again
         # Also handle malformed bibtex and metadata
+
+    rp.close()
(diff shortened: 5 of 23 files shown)


omangin:
Great! It indeed is fast (again). I have refactored your code a little bit to remove code duplication and added some unittests for the core cache functionalities (that are now separate). It seems to work now. I still need to play around a little more with it but then we will be ready to merge.

Just a design concern: the fact that `pull_cache` is a method of `databroker` is not straightforward for me. Maybe the cache storage format is cache specific and thus should only appear in cache related code. Similarly I am not sure the `filebroker` should create the cache directories by default. I think this kind of break the design of having the cache completely optional. What is your opinion?

Finally I was wondering why the creation of the `databroker` is delayed in the `datacache`?

Thanks for the great work, though!

humm:
> Just a design concern: the fact that pull_cache is a method of databroker is not straightforward for me. Maybe the cache storage format is cache specific and thus should only appear in cache related code. Similarly I am not sure the filebroker should create the cache directories by default. I think this kind of break the design of having the cache completely optional. What is your opinion?

Well, creating directories just-in-time, just-as-needed, seemed it would complexify the code. I don't see the cache as optional. It is a feature, and a behaviour of the code. I don't expect to let users disable the cache system, nor them ever bothering about it.

In that perspective, databroker handle data, not necessarily only the bib and meta data. Similarly, filebroker is our entry point to the disk, and I don't see a reason to duplicate the code to isolate the cache. I get that isolating the cache code might seem cleaner, but it seems that it would mean maintaining two very overlapping systems. 

> Finally I was wondering why the creation of the databroker is delayed in the datacache?
No real reason. The pattern was already present, I reused it. Only creating the resource when needed was a design idea, but there is no practical point with the code as it is. So, yeah, there is some room to simplify the code. 

BTW, after using it a bit, there are some issues with the cache. Editing a bibtex will rewrite the entire bibcache, which takes ~1s on my repo, and makes pubs hang after the edits are made. Obviously, this is not ideal. I guess, cache should not be written after all commands, but maybe only general ones, such as list. I'll get around to it in a few days.

Nice code factoring. I had a nagging voice in my head saying that was writing too much of the same code while doing this, but given all the other problems, I did not get around to listen to it.

omangin:
> In that perspective, databroker handle data, not necessarily only the bib and meta data. Similarly, filebroker is our entry point to the disk, and I don't see a reason to duplicate the code to isolate the cache. I get that isolating the cache code might seem cleaner, but it seems that it would mean maintaining two very overlapping systems.

Fair enough. I was actually only thinking of moving the functions `push_cache` and `pull_cache` to the `datacache`.

> BTW, after using it a bit, there are some issues with the cache. Editing a bibtex will rewrite the entire bibcache, which takes ~1s on my repo, and makes pubs hang after the edits are made. Obviously, this is not ideal. I guess, cache should not be written after all commands, but maybe only general ones, such as list. I'll get around to it in a few days.

I'd say that what is surprising is that it takes one second to write the cache. It seems that if there is no way of improving that, we have to re-think the whole design of the cache system.

humm:
> Fair enough. I was actually only thinking of moving the functions `push_cache` and `pull_cache` to the datacache.

I see. The thing I wanted to avoid was datacache having direct access to the filebroker. I view datacache as a buffer zone between the commands and the databroker. It having direct access to the filesystem seemed messy, design-wise.

> I'd say that what is surprising is that it takes one second to write the cache. It seems that if there is no way of improving that, we have to re-think the whole design of the cache system. Maybe we need to do some benchmarking here. Is the problem occurring with both python 2 and 3?

Yes, benchmarking would be nice. We could use something like [airspeed velocity](https://github.com/spacetelescope/asv). I will look into this hanging thing. I noticed it only once or twice.

Issue #58, opened 2017-04-14
Broken tests for python2


omangin:
The use of open instead of io.open breaks the compatibility with python2.

benureau:
Ok, so I upgraded pyfakefs from 2.8 to 3.3(dev) and some work was needed to make our test suite work again (see [bug/unittests](/pubs/pubs/tree/bug/unittests)). I'll be attacking the compatibility with Python 2 next.

benureau:
Resolved with #73

PR #59, opened 2017-05-05
fixed inverted conditional

tmearnest:


omangin:
Thanks @tmearnest. However, I don't really understand your change. The encoding option for open only exists for Python 3 and is only valid for text mode, not binary. What is the change supposed to fix?

Overall I believe this is related to issue #58 because we should be using `io.open` for compatibility between Python 2 and 3. @benureau, did you have a look at the issue?

tmearnest:
Sorry, I guess I was a bit terse. On Python 3, the previous code will throw

    ValueError: binary mode doesn't take an encoding argument

since `open(system_path(path), mode, encoding='utf-8')` will be called if  `'b' in mode`. 

Your original code tested for `mode.find('b') == -1` which is equilivent to `not ('b' in mode)`

omangin:
Indeed, you're right. Sorry, for that. I'll merge the pull request, then.

Issue #34, opened 2015-12-03
Autocompletion for citekeys


73:
Hi Guys,

I would like suggest autocompletion for (well that's what I would like) zsh.
This might be a place to start: https://github.com/zsh-users/zsh-completions

Let me know what you think.
--73

humm:
Yes, autocompletion would be nice. For the moment, we are looking into basic command autocompletion using https://github.com/kislyuk/argcomplete, which is zsh compatible. but citekey autocompletion is the next step.

humm:
FYI, I looked and tried to add autocompletion with argcomplete, but it is dicier than it looks. Given the current code, argcomplete is not a good option.

73:
Thank you for looking l. I really appreciate that. What is the problem?

humm:
Besides the fact that I did not actually manage to make it work, argcomplete actually execute the program up until parse_args(). I guess our double parser solution is wreaking havoc in argcomplete, but it also means that the conf needs to be loaded everytime the user hits tab, which is way too slow and produce noticeable delays in the terminal.

73:
Ah okay. If the commands are stable we could write a auto complete function
shell only. Or is there another option?

humm:
IDK

omangin:
An issue will be that with a lot of papers it is very long to parse all files. This will be solved by the cache but with the current implementation `list` reads all files even when it just has to output the citekeys.

omangin:
With the current implementation `list` reads all files even when it just has to output the citekeys. This can take a long time with a few hundreds of papers. This will be solved by the cache.  In the meanwhile it is also possible to hack list to not build the full repository when only asked for citekeys but this adds complexity to the code that will become useless in the future.

73:
Since we are talking completion here, the fastest way would be to skip the `list`command and act on a listing of the `bib ` folder with trimmed file endings. One would still need to read the config to determine the location of the folder. Anyway, in my opinion it would make `pubs`a lot more usable if i would not have to type (which on rather long keys means copy & paste) the specific keys.  
--73

humm:
Let's remark that typing part of the citekey *will* autocomplete implicitly if only one correspond to the prefix, and ask for disambiguations with a list of options if not. This is the case for the open command for instance (we may not be the case for all commands). This is *very* close to what you want.

73:
True. Open fully supports it. Haven't tasted it though. Will test it. Maybe
that will suffice. Will get back to that.
--73

[...]

PR #61, opened 2017-05-13
Feat/autocomplete

omangin:
Basic implementation of shell autocompletion with argcomplete (see #34). Any testing, suggestion, or comment is welcome.

Note: the tests are failing on python2 due to #58.

### DIFF
--- a/pubs/commands/add_cmd.py
+++ b/pubs/commands/add_cmd.py
@@ -9,7 +9,7 @@ from .. import color
 from .. import pretty
 
 
-def parser(subparsers):
+def parser(subparsers, conf):
     parser = subparsers.add_parser('add', help='add a paper to the repository')
     parser.add_argument('bibfile', nargs='?', default=None,
                         help='bibtex file')
--- a/pubs/commands/conf_cmd.py
+++ b/pubs/commands/conf_cmd.py
@@ -17,7 +17,7 @@ def command(conf, args):
         # get modif from user
         ui.edit_file(config.get_confpath())
 
-        new_conf = config.load_conf(check=False)
+        new_conf = config.load_conf()
         try:
             config.check_conf(new_conf)
             ui.message('The configuration file was updated.')
--- a/pubs/commands/conf_cmd.py
+++ b/pubs/commands/conf_cmd.py
@@ -3,7 +3,7 @@ from .. import config
 from .. import content
 
 
-def parser(subparsers):
+def parser(subparsers, conf):
     parser = subparsers.add_parser('conf',
             help='open the configuration in an editor')
     return parser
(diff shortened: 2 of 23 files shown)


benureau:
So I tried to make it work, but there are a few issues on macOS, one of them being that bash is outdated. I have upgraded bash, made it the login shell, and some other voodoo stuff, and I have made it work. On the plus side, having it not work does not seem to affect the other behavior of pubs so that's that. I'll add a few words for macOS users in the readme.

The command `open` does not seem to have completion enabled (I have not look into the code yet). It is slow: can it be avoided by using caches?

omangin:
Great! The open command does not exist anymore and you probably have it set as an alias to `doc open`. Aliases are not supported yet so that's why it does not work.

Regarding the speed, the completion already uses the cache. On my computer with 500 papers it's quite OK though (<0.5s). The eventual delay comes from
– what is loaded before [the `autocomplete` command](https://github.com/pubs/pubs/blob/feat/autocomplete/pubs/pubs_cmd.py#L86). I'm not sure we have such a big margin to optimize there,
– the command to get the list of citekeys which currently lists the content of the directory (see the `databroker`). This could be replaced by a cached list of citekeys and a check to the directory last modified date but I'm not sure it would save much time.
In any case we should profile it first.
The tag completion is slower because it requires loading the metatdata from all papers in the repository. We could implement caching the tags at directory level but it would still require (if we want to keep accuracy) that all metadata are checked for update. Once again, I'm not sure it worth it.

benureau:
Hum... I have 900 papers and it's more like ~1s on mine (SSD and all). I'll look into it, I feel that there are some low-hanging fruits here on speed. But it's a different issue. I'll try to add the setup instructions soon.

omangin:
> I'll try to add the setup instructions soon.
There are already some in the `readme` but they might benefit some additions.

omangin:
BTW, here are the times I get for loading the main components of pubs in `ipython` with `%time`:
- 47 ms for `from pubs.ui import get_ui`
- 2 ms for `get_ui()`
- 13 us for `from pubs import config`
- 12 us for `config.get_confpath(verify=False)`
- 778 us for `load_conf(conf_path)`
- 90 us for `repo.Repo(conf)`
- 3 ms for `rp.citekeys`
- 85 ms for `rp.get_tags`

So it does not seem that any of them is critically slow  (thanks to the cache for the last two, which as a matter of fact are the only ones that should scale with the size of the repository). The overall time seems to match what I experience (with some additional `argparse` and `argcomplete` overhead). 

@benureau Maybe you could check if you get the same times on your machine.

If we cannot solve the speed this time and it's the only issue with the PR, I'd suggest we merge anyway, mark completion as experimental, and open a new issue to get more feedback.

benureau:
Ok. Created #72 to deal with the speed. Merging this.

PR #62, opened 2017-05-24
default message for errors bugfix

d9w:
There is a bug in `pubs/repo.py`, which references the `default_msg` of the error classes instead of their `default_message`


omangin:
Indeed. Thanks!

Issue #57, opened 2016-08-09
Cannot edit conf


minthemiddle:
When running `pubs conf` (both as user and root in a virtualenv, latest Win and OSX), I get an error message:

`error: [Errno 13] Permission denied`

humm:
Hum... your conf file should be located at `~/.pubsrc`. Can you edit it directly, with, for instance `vim ~/.pubsrc`? By the way, we tested pubs on OSX and Linux, but never on Windows. You may get strange issues there.

omangin:
I assume this was fixed at the same time as #63. Please reopen if still an issue.

PR #63, opened 2017-05-24
use vim as an editor command by default

d9w:
This is a potential fix for #57, which arises from a call to the default editor and then a temporary file. As the default editor is the empty string, this results in `subprocess.call(['/tmp/tempfile'])`, which errors `error: [Errno 13] Permission denied` as the file is not executable.

Using vim as the default is not a perfect solution, but at least the error message will now be more clear if the user does not have vim installed: `error: [Errno 2] No such file or directory: 'vim'`

### DIFF
--- a/pubs/config/spec.py
+++ b/pubs/config/spec.py
@@ -20,7 +20,7 @@ open_cmd = string(default=None)
 # if using a graphical editor, use the --wait or --block option, i.e.:
 # "atom --wait"
 # "kate --block"
-edit_cmd = string(default='')
+edit_cmd = string(default='vim')
 
 # If true debug mode is on which means exceptions are not catched and
 # the full python stack is printed.


omangin:
Thanks @d9w. The bug is actually that the configuration value should not be used to provide a default editor. So the `content.edit_file(conf['main']['edit_cmd'], …)` should never be called in the first place. Instead `ui.editor` should be used as the command because it falls back to the `$EDITOR` defined as an environment variable or to nano if not set (we chose nano over vi but we can switch if that's preferred; most linux systems have the environment variable set anyway).

I'll fix that later today. Thanks anyway for the PR.

omangin:
So, I fixed the issue by using the UI functions and moved the ones from `content` away to avoid confusion.

Thinking about it, I'm not sure that this is the best solution but at least it's consistent now. The empty string for the editor in the configuration means _undefined_ and falls back to to the system wide setting (through the environment variable).

PR #64, opened 2017-05-24
link by default and allow overwriting when importing a bibtex file

d9w:
I personally prefer overwriting existing entries when importing a bibtex file, so I created a flag `-O` or `--overwrite` to allow that and set it to `True` by default.

In doing so, I also pass the copy config value to `push_doc`. If `overwrite` is `True` and the `copy` configuration is `copy` or `move`, the import will fail as the file already exists. For this reason, and because I think `link` is a more reasonable default (doesn't bloat a `~/.` directory and doesn't move pdfs that are maybe being used by other programs, like Mendeley), I also changed the default to `link`.

### DIFF
--- a/pubs/commands/import_cmd.py
+++ b/pubs/commands/import_cmd.py
@@ -20,6 +20,8 @@ def parser(subparsers):
             help="don't copy document files, just create a link.")
     parser.add_argument('keys', nargs='*',
             help="one or several keys to import from the file")
+    parser.add_argument('-O', '--overwrite', default=True,
+            help="Overwrite keys already in the database")
     return parser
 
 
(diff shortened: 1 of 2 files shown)


omangin:
Thanks @d9w. That's kind of two things in one, so let's take them in order.

Regarding the overwrite option, that's indeed a very good thing to have. I'd suggest however that we stick with the current default. The reason is that common citekey schemes allow collisions (e.g. `name + year + [abcd...]`), hence import might then overwrite a paper by another distinct paper that has the same citekey. I'd rather keep a conservative approach to that. We can nevertheless add a configuration option like we have for the copy/move/link case. On the long run, this connects to a question about the ability to interactively merge two bibtex files about the same entity, including which document to keep if one already exists. I'll open an issue to discuss this.

For the default on linking rather than moving, I believe that when overwrite is set, the sane behavior is to also overwrite the document. I believe that it's more in the spirit of pubs  to store everything at the same place (and hence having _copy_ or _move_ as the default), but I'm happy to discuss that point too if needed.

As a summary I'd suggest that:
– we open a separate issue to discuss improvements on merging / overwriting,
– you make this PR specific to the _overwrite_ option (moving away the part on the `doc_add` configuration),
– you eventually open an issue to discuss the default values of both options.

d9w:
Sounds good. Closing this PR and opening a new one for now.

PR #66, opened 2017-05-29
overwrite argument on import

d9w:
This allows for overwriting existing entries based on argument input for the `import` command. Note that this will currently fail when overwriting an entry with a `file` field and existing doc, due to line `89:                rp.push_doc(p.citekey, docfile, copy=copy)`. See #64 for more discussion.

### DIFF
--- a/pubs/commands/import_cmd.py
+++ b/pubs/commands/import_cmd.py
@@ -20,6 +20,8 @@ def parser(subparsers):
             help="don't copy document files, just create a link.")
     parser.add_argument('keys', nargs='*',
             help="one or several keys to import from the file")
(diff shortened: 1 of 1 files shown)


omangin:
Looks good! For the document, it would make sense to also overwrite it when the option is set. It should actually already happen when we pass the option to `push_paper`. I'll have a look on how to fix that when I've a little more time.

The other thing to do before merging is to add a test for when the option is set in `tests/test_usecase.py`. Testing that it does not fail and does not increment the number of papers. Can you have a look at that?

We could also add a similar option for the `add` command. But that does not need to be part of this PR.

omangin:
Merged.

PR #67, opened 2017-05-29
Keylist

d9w:
This feature allows the user to configure a list of keys for the exported bib. For example, I dislike having the `abstract` text in my `bib` files, so I've configured my keylist as:
`keylist = author, title, journal, publisher, year, month, number, volume, pages, url`

An empty keylist, the default, maintains current behavior.
Sorry if this is a lot of pull requests ^.^

### DIFF
--- a/pubs/endecoder.py
+++ b/pubs/endecoder.py
@@ -106,7 +106,7 @@ class EnDecoder(object):
                 bibraw += '    {} = {{{}}},\n'.format(
                     key, EnDecoder._encode_field(key, value))
         for key, value in bibentry.items():
-            if key != TYPE_KEY:
+            if key != TYPE_KEY and (not keylist or key in keylist):
                 bibraw += '    {} = {{{}}},\n'.format(
                     key, EnDecoder._encode_field(key, value))
         bibraw += '}\n'
--- a/pubs/endecoder.py
+++ b/pubs/endecoder.py
@@ -76,9 +76,9 @@ class EnDecoder(object):
     def decode_metadata(self, metadata_raw):
         return yaml.safe_load(metadata_raw)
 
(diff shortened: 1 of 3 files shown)


pubs/endecoder.py
```diff
@@ -97,16 +97,16 @@ def _encode_field(key, value):
             return value
 
     @staticmethod
-    def _encode_bibentry(citekey, bibentry):
+    def _encode_bibentry(citekey, bibentry, keylist):
         bibraw = '@{}{{{},\n'.format(bibentry[TYPE_KEY], citekey)
         bibentry = copy.copy(bibentry)
         for key in bibfield_order:
-            if key in bibentry:
+            if key in bibentry and (not keylist or key in keylist):
```
omangin: Is the `not keylist` still needed after your latest commit?

pubs/endecoder.py
```diff
@@ -97,16 +97,16 @@ def _encode_field(key, value):
...
                 value = bibentry.pop(key)
                 bibraw += '    {} = {{{}}},\n'.format(
                     key, EnDecoder._encode_field(key, value))
         for key, value in bibentry.items():
-            if key != TYPE_KEY:
+            if key != TYPE_KEY and (not keylist or key in keylist):
```
omangin: Same as L104. Maybe it is now also worth refactoring the two in a function.

omangin:
That seems a good idea! I'd just change the option name to `exclude-fields` to make it more explicit. _Key_ might be confusing because we use it a lot for citekeys.

After that we might consider adding a few tests first at the `endecoder` level and then in `test_usecase`.

omangin:
@d9w, any chance to look at the comment?

PR #73, opened 2017-07-11
Bug/unittests

benureau:


pubs/content.py
```diff
@@ -70,9 +70,14 @@ def read_text_file(filepath, fail=True):
     try:
         with _open(filepath, 'r') as f:
             content = f.read()
+            try:  # Python 2
```
omangin: We could move this one to the `_open` function so that all python 2 stuff is handled there. It would return a `bytestring` only when `'b'` is in the `mode` argument, regardless of python version.

pubs/content.py
```diff
@@ -47,12 +48,11 @@ def system_path(path):
 
 
 def _open(path, mode):
-    if 'b' in mode:
+    if 'b' in mode or sys.version_info < (3,):
```
omangin: Should we move this one to the `p3` module?

pubs/uis.py
```diff
@@ -14,6 +14,7 @@
 from .content import check_file, read_text_file, write_file, system_path
 
 
+DEBUG = False
```
omangin: Do we need to keep this one?

pubs/uis.py
```diff
@@ -117,7 +118,7 @@ def handle_exception(self, exc):
 
         :returns: True if exception has been handled (currently never happens)
         """
-        if not self.debug:
+        if (not DEBUG) and (not self.debug):
```
omangin: Same as above.

setup.py
```diff
@@ -19,7 +19,8 @@
                             'pubs.plugs', 'pubs.plugs.alias'],
     scripts      = ['pubs/pubs'],
 
-    install_requires = ['pyyaml', 'bibtexparser', 'python-dateutil', 'requests', 'configobj',
+    install_requires = ['pyyaml', 'bibtexparser', 'python-dateutil', 'requests',
+                        'configobj', 'argcomplete',
```
omangin: (This commit should not be part of this PR.)
Anyway, since the auto-completion still requires manual configuration I don't believe it worth it to include it as a requirement. We should however definitely add it as an `extras_require`.

tests/fake_env.py
```diff
@@ -5,8 +5,11 @@
 import glob
 
 import dotdot
-from pyfakefs import (fake_filesystem, fake_filesystem_shutil,
-                      fake_filesystem_glob, fake_filesystem_unittest)
+
+from pyfakefs import fake_filesystem, fake_filesystem_unittest
+
+# from pyfakefs import (fake_filesystem, fake_filesystem_shutil,
+#                       fake_filesystem_glob, fake_filesystem_unittest)
```
omangin: The commented lines can be removed.

tests/test_usecase.py
```diff
@@ -124,7 +124,7 @@ def execute_cmds(self, cmds, capture_output=CAPTURE_OUTPUT):
                     self.fail('Unexpected input asked by command: {}.'.format(
                         actual_cmd))
             if PRINT_OUTPUT:
-                print(outs)
+               print(outs)
```
omangin: Missing space

omangin:
So, I removed the commit with the requirements and fixed the indentation, an unused import, and the commented code.

Feel free to remove the `DEBUG` in `uis.py` Otherwise, the code is good to merge for me!

pubs/uis.py
benureau: Yep. There is no easy way to make exceptions verbose in the usecase tests, since it involves changing the configuration file. We need a way to make gui error message into true exception when debugging test: this is a simple way to do that.

pubs/content.py
benureau: I don't know if that would make it any clearer. If you have an idea, I am all hear.

benureau:
@omangin Thanks for the review and the changes! Merging.

PR #74, opened 2017-07-11
(Optional) Requirement

omangin:
@benureau I leave it to you to decide whether it should be optional or not (I'm in favor of the former, though).

Same thing for `beautifulsoup` (for which I have no opinion since the functionality does not require configuration). The change would require some fallback in `apis.py`.


benureau:
Yes, I had a feeling you would make this remark about requiring `argcomplete`. You got a good point though.

Issue #76, opened 2017-07-12
Cannot add DOI because "error: 'ascii' codec can't encode character ..."


eschen42:
I installed pubs with
```
pip install --upgrade git+https://github.com/pubs/pubs
```
Should the `--upgrade` switch have been included in the instructions on the README.md page?

I have a DOI that apparently pubs add is allergic to (another DOI added fine):
```
$ pubs add -D 10.1021/ac0713510
error: 'ascii' codec can't encode character u'\xf6' in position 80: ordinal not in range(128)
```
One of the authors is named `Sjöström`; position 80 is that of the first vowel in this name.

After this, 
```./bib/Wiklund_2008.bib```
exists with zero size, and I cannot use
```pubs remove Wiklund_2008```
to get rid of it; hence, if I keep trying I accumulate Wiklund_2008.bib, Wiklund_2008a.bib, Wiklund_2008b.bib, Wiklund_2008c.bib etc., all with zero size.

I tried to help myself without success:
```
$ pubs conf
error: edit_file() takes exactly 3 arguments (2 given)
```

I gather from [issue 51](https://github.com/pubs/pubs/issues/51) that I  need to install the `utf-8` codec
but I do not see instructions regarding how to do so.

benureau:
Thanks for the issue!

The `--upgrade` is needed if you already installed `pubs`, to get the lasted version.

I was able to reproduce your bug. It appears only with Python 2, with Python 3 you don't have the problem. But switching to python 3 might not be an option for you. 

I will attend the bug, but not before Friday, unless @omangin gets to it first.

eschen42:
Awesome. Thank you.

omangin:
Thanks @eschen42 for the notice.

There were actually several issues here.

- an actual bug on writing non-ascii content from python 2 (which ended up leaving an empty file as you told),

- this error was not caught by the `add` command,

- the remove command does not work when there is no metadata for the paper. This comes from the fact that we use metadata files as the source for the citekeys in the repository. Hence the `add` command failing to write the metadata file, the repository was in a broken state. This specific case could be solved by switching the following lines in the `remove` method from the repository:
  ````
   self.citekeys.remove(citekey)
   self.databroker.remove(citekey)
   ````
  We could also check if the citekey exists in the repository, raise a warning if not and proceed to remove it anyway. I'm however not in favor of this, and I would rather implement some repository sanity check instead with some command to interactively clean "orphaned" bib files or create metadata files for them. Any comment?

[...]

PR #77, opened 2017-07-14
Fix for #76

omangin:
There is actually a little more than the fix (see commit messages for more details).

Note: I didn't add any non-regression test at the usecase level because these things are not really caught by the `fakefilesystem` thing. It might worth spending a little time checking for these or eventually adding a few assertions directly in the `write_file` method from `content.py` to make sure that it's called properly (i.e. no binary content without binary mode). The latter is probably the quickest way to check that.

### DIFF
--- a/pubs/content.py
+++ b/pubs/content.py
@@ -94,7 +94,16 @@ def remove_file(filepath):
 
 
 def write_file(filepath, data, mode='w'):
+    """Write data to file.
+
+    Data should be unicode except when binary mode is selected,
+    in which case data is expected to be binary.
+    """
     check_directory(os.path.dirname(filepath))
+    if 'b' not in mode and sys.version_info < (3,):
+        # _open returns in binary mode for python2
+        # Data must be encoded
+        data = data.encode('utf-8')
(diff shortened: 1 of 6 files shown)


pubs/content.py
```diff
@@ -52,16 +52,17 @@ def _open(path, mode):
     else:
```
benureau: I like 'Cosmit' a lot.

omangin: I should have called it PEP8 ;)

Is this comment related in any way to the line it's attached to?

benureau: No no no. To the title of you commit. But I did not find any way to comment that directly.

benureau:
@omangin: I was wondering about adding things like DOI tests in the usecases, but that would mean the test would depend on access to the internet. What do you think?

omangin:
> @omangin: I was wondering about adding things like DOI tests in the usecases, but that would mean the test would depend on access to the internet. What do you think?
 
I was worried about the tests needing a connection to work but I think it's worth it to have the tests, so that we'd also be noticed if the APIs change. I'll add a few tests before merging.

PR #79, opened 2017-07-17
Fixes websearch in Python 3 (fix #69)

### LINKED ISSUE #69: websearch fails for python3
If run with `python3` the `websearch` command fails:

```bash
$ python --version
Python 3.6.1
$ pubs websearch <string>
error: module 'urllib' has no attribute 'quote_plus'
```

`urllib` has been split up into different subpackages, and the `quote_plus` function is now in `urllib.parse`. Thus `urllib.quote_plus` has to be changed to `urllib.parse.quote_plus` in `websearch_cmd.py`, however this will break `python2` compatibility?

benureau:
Simple fix, should not interfere much with other PRs.

### DIFF
--- a/pubs/commands/websearch_cmd.py
+++ b/pubs/commands/websearch_cmd.py
@@ -1,6 +1,6 @@
 import webbrowser
-import urllib
 
+from .. import p3
 from ..uis import get_ui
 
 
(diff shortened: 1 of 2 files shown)


benureau:
Yes, a testcase is not possible directly.

PR #82, opened 2017-07-22
Bug/conf cmd

benureau:
* `pubs conf` was broken. Fixed that and added a regression test.

### DIFF
--- a/tests/test_usecase.py
+++ b/tests/test_usecase.py
@@ -536,7 +536,8 @@ class TestUsecase(DataCommandTestCase):
                 ('pubs conf', [str_fixtures.sample_conf]),
                ]
         self.execute_cmds(cmds)
-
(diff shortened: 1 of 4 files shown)


tests/test_usecase.py
```diff
@@ -531,6 +531,13 @@ def clean(s):
         self.assertEqual(clean(correct[2]), clean(out[2]))
         self.assertEqual(clean(correct[4]), clean(out[4]))
 
+    def test_conf(self):
+        cmds = ['pubs init',
+                ('pubs conf', [str_fixtures.sample_conf]),
+               ]
+        self.execute_cmds(cmds)
```
omangin: While you are there, maybe you can check that the configuration file is updated by the command. Otherwise the test should be called `test_conf_does_not_fail_without_any_reason` 😉 !

PR #91, opened 2017-09-28
[Fix #88] Adds proper escaping for arguments in alias plugin.

### LINKED ISSUE #88: "pubs list" has different behavior in custom commands than in terminal
In a terminal, if you use the command `pubs list -k abstract:"Only a few genes remain"` to retrieve the key corresponding to the article `doi:10.1371/journal.pgen.1002876`, it indeed outputs you the right key.

However, if you define the command `exportlist = !pubs list -k "$@"` in the conf file and try `pubs exportlist abstract:"Only a few genes remain"`, the output is `error: Invalid query (a)`.

omangin:


pubs/p3.py
```diff
@@ -52,6 +54,7 @@ def _get_fake_stdio_ucontent(stdio):
         return stdio.read()
 
     import pickle
+    from pipes import quote as shell_quote
```
benureau: I don't get why this line is repeated and not put outside of the if/else if it is the same for both versions.

benureau: Is it relevant in p3.py by the way?

omangin: Well spotted! The py3 version of the line should be importing `quote` from `shlex`…

Issue #99, opened 2017-11-28
Print help on when no subcommands are supplied


wflynny:
When attempting to use a less-often-used command, it'd be nice if `pubs` printed the main parser help.  For example, this is how the git CLI works.  Are the maintainers open to this behavior?  I've got a [commit on a fork](https://github.com/pubs/pubs/compare/master...wflynny:help-when-no-subcommand) that does just this.

omangin:
Yes, that makes sense to me. Can you create a pull request for your commit? Also the unit test you removed should be instead updated to check for the expected behavior.

Thanks @wflynny!

PR #100, opened 2017-11-30
Print help when no subcommand is provided (#99)

wflynny:
This PR implements functionality discussed in Issue #99 for `pubs` to be called without a subcommand and it will print the help menu and exit.  This is similar behavior to how the git CLI functions.  A unittest has been modified to cover this behavior, though the test coverage isn't great due to difficulties with argparse's handing of the `-h/--help` flags.

### DIFF
--- a/pubs/pubs_cmd.py
+++ b/pubs/pubs_cmd.py
@@ -83,13 +83,17 @@ def execute(raw_args=sys.argv):
 
         # Eventually autocomplete
         autocomplete(parser)
+
         # Parse and run appropriate command
+        # if no command, print help and exit peacefully (as '--help' does)
         args = parser.parse_args(remaining_args)
+
         if not args.command:
             parser.print_help()
-        else:
-            args.prog = "pubs"  # FIXME?
-            args.func(conf, args)
+            sys.exit(0)
+
+        args.prog = "pubs"  # FIXME?
+        args.func(conf, args)
 
     except Exception as e:
         if not uis.get_ui().handle_exception(e):
--- a/pubs/pubs_cmd.py
+++ b/pubs/pubs_cmd.py
@@ -69,7 +69,7 @@ def execute(raw_args=sys.argv):
                                          prog="pubs", add_help=True)
         parser.add_argument('--version', action='version', version=__version__)
         subparsers = parser.add_subparsers(title="valid commands", dest="command")
-        subparsers.required = True
+        #subparsers.required = True
 
         # Populate the parser with core commands
         for cmd_name, cmd_mod in CORE_CMDS.items():
@@ -85,8 +85,11 @@ def execute(raw_args=sys.argv):
         autocomplete(parser)
         # Parse and run appropriate command
         args = parser.parse_args(remaining_args)
-        args.prog = "pubs"  # FIXME?
-        args.func(conf, args)
+        if not args.command:
+            parser.print_help()
+        else:
+            args.prog = "pubs"  # FIXME?
+            args.func(conf, args)
 
     except Exception as e:
         if not uis.get_ui().handle_exception(e):
--- a/pubs/pubs_cmd.py
+++ b/pubs/pubs_cmd.py
@@ -69,7 +69,6 @@ def execute(raw_args=sys.argv):
                                          prog="pubs", add_help=True)
         parser.add_argument('--version', action='version', version=__version__)
         subparsers = parser.add_subparsers(title="valid commands", dest="command")
-        #subparsers.required = True
 
         # Populate the parser with core commands
         for cmd_name, cmd_mod in CORE_CMDS.items():
@@ -86,11 +85,11 @@ def execute(raw_args=sys.argv):
 
         # Parse and run appropriate command
         # if no command, print help and exit peacefully (as '--help' does)
-        args = parser.parse_args(remaining_args)
-
+        args = parser.parse_args(remaining_args) 
         if not args.command:
-            parser.print_help()
-            sys.exit(0)
+            ui.error("Too few arguments!\n")
+            parser.print_help(file=sys.stderr)
+            sys.exit(2)
 
         args.prog = "pubs"  # FIXME?
         args.func(conf, args)
--- a/tests/test_usecase.py
+++ b/tests/test_usecase.py
@@ -187,13 +187,6 @@ class URLContentTestCase(DataCommandTestCase):
 
 # Actual tests
 
-class TestAlone(CommandTestCase):
-
-    def test_alone_misses_command(self):
-        with self.assertRaises(FakeSystemExit):
-            self.execute_cmds(['pubs'])
-
-
 class TestInit(CommandTestCase):
 
     def test_init(self):
--- a/tests/test_usecase.py
+++ b/tests/test_usecase.py
@@ -29,14 +29,21 @@ PRINT_OUTPUT   = False
 CAPTURE_OUTPUT = True
 
 
-class FakeSystemExit(SystemExit):
+class FakeSystemExit(Exception):
     """\
     SystemExit exceptions are replaced by FakeSystemExit in the execute_cmds()
     function, so they can be catched by ExpectedFailure tests in Python 2.x.
 
     If a code is expected to raise SystemExit, catch FakeSystemExit instead.
+
+    Added explicit __init__ so SystemExit.code functionality could be emulated.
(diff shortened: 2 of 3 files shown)


wflynny:
Hmm, I'll open up a 2.7 and 3.3 env later and see if I can resolve these issues.

wflynny:
After looking into the 2.7 failure, it seems like this entire enhancement (optional subparsers) is not possible in 2.7 ([ref](https://stackoverflow.com/questions/22990977/why-does-this-argparse-code-behave-differently-between-python-2-and-3)).  In that case, we may need to reconsider the entire feature.  

Should I focus on making it 3.3 only?  Or should I abandon this idea?

pubs/pubs_cmd.py
```diff
@@ -83,8 +83,15 @@ def execute(raw_args=sys.argv):
...
         # Eventually autocomplete
         autocomplete(parser)
+
         # Parse and run appropriate command
+        # if no command, print help and exit peacefully (as '--help' does)
         args = parser.parse_args(remaining_args)
+
+        if not args.command:
+            parser.print_help()
+            sys.exit(0)
```
omangin: Here we would need to return an exit code `2` to be consistent with `argparse` and because this is still an error.

tests/test_usecase.py
```diff
@@ -29,7 +29,7 @@
 CAPTURE_OUTPUT = True
 
 
-class FakeSystemExit(Exception):
+class FakeSystemExit(SystemExit):
```
omangin: Is this needed? Let's not start a philosophical debate about whether or not a `FakeSystemExit` is a `SystemExit` but in the new version it would not be cached by an `except Exception` which is probably not wanted (although `unittest` seems to handle that correctly).

tests/test_usecase.py
```diff
@@ -131,9 +131,9 @@ def execute_cmds(self, cmds, capture_output=CAPTURE_OUTPUT):
             exc_class, exc, tb = sys.exc_info()
             if sys.version_info.major == 2:
                 # using six to avoid a SyntaxError in Python 3.x
-                six.reraise(FakeSystemExit, exc, tb)
+                six.reraise(FakeSystemExit, *exc.args, exc_traceback=tb)
```
omangin: I think this line should not be changed. Am I wrong?

pubs/pubs_cmd.py
wflynny: `argparse._HelpAction.__call__` prints to stderr but then calls `argparse.ArgumentParser.exit()` which defaults to just a `sys.exit(0)`.  So when comparing exit codes of the `SystemExit`s that are raised, I thought this should be zero.  Is that wrong?

omangin: What I meant is that not providing arguments should probably be considered an error (e.g. `git` returns with code `1` in the same situation). This is mainly to be consistent, so that a scripts _knows_ that the status is a failure.

On the other side we could consider that showing help is the default behavior and is not an error, but I think the other behavior is more consistent.

wflynny: Yep, you're right.  So do you think it should be error code `1` (like git) or `2`.  `1` seems more appropriate (operation not permitted).

omangin:
Thanks for the PR. It seems that the feature is not as simple to implement as it initially looked.

Considering that this is really a minor feature (actually more on the documentation side), I would be in favor of targeting an implementation as simple as possible.

Therefore, although the _right_ way to do this would be to subclass the argument parser, this is way too complex for what we want (thanks to `argparse` being poorly designed).

For me, the important feature (the one that should stay in the test) is that the code exits with code `2` when the argument list is incomplete. Hence, we can keep the unit test that was there (and that's the only mandatory thing to test).

Because what this PR brings is more about documentation than feature, it is OK for me if the printed text is not the same on python 2 and 3 (that is, the regular message in py2 and the error message plus the full help message in py 3). 

It's also OK to drop the PR altogether if you prefer. I'll complete a review with what needs to be changed if you want to finish this. If not we would need to PR https://github.com/pubs/pubs/pull/100/files#diff-cccc3c54ebff14162cb6e8fab046309cL136 anyway because it is fixing a bug.

@benureau, any additional thought?

omangin:
Sorry I didn't mean to close it.

tests/test_usecase.py
wflynny: The code in the current PR is wrong for Python2 as `six.reraise` looks like its arguments have been renamed moving to Python3.   `six.reraise(FakeSystemExit, *exc.args, tb=tb)` works. 

The reason for this change is dependent on the change of `FakeSystemExit`s base class.  Its so the `code` attribute is set to `code=0` and not `code=SystemExit(0,)`.  It could probably be changed to `reraise(FakeSystemExit, exc.args[0], tb)`.

pubs/pubs_cmd.py
```diff
@@ -69,7 +69,7 @@ def execute(raw_args=sys.argv):
                                          prog="pubs", add_help=True)
         parser.add_argument('--version', action='version', version=__version__)
         subparsers = parser.add_subparsers(title="valid commands", dest="command")
-        subparsers.required = True
+        #subparsers.required = True
```
omangin: Remove the line altogether.

tests/test_usecase.py
wflynny: Base `Exceptions` don't have the `code` attribute like `SystemExit`s do.  I can change it to subclass `Exception` and redefine `__init__`.  This just seemed simpler and didn't break the unittests (for now at least).

pubs/pubs_cmd.py
```diff
@@ -83,8 +83,15 @@ def execute(raw_args=sys.argv):
 
         # Eventually autocomplete
         autocomplete(parser)
+
         # Parse and run appropriate command
+        # if no command, print help and exit peacefully (as '--help' does)
         args = parser.parse_args(remaining_args)
+
+        if not args.command:
```
omangin: We need to also print the error message ("too few arguments") to `stderr`.

pubs/pubs_cmd.py
```diff
@@ -83,8 +83,15 @@ def execute(raw_args=sys.argv):
...
+            parser.print_help()
```
omangin: This shoud probably go to `stderr` too. (Use `parser.print_help(file=...)`.)

pubs/pubs_cmd.py
omangin: `2` is consistent with what we have for all the other argument errors.

tests/test_usecase.py
omangin: Probably remove this.

tests/test_usecase.py
```diff
@@ -131,9 +131,9 @@ def execute_cmds(self, cmds, capture_output=CAPTURE_OUTPUT):
...
             else:
-                raise FakeSystemExit(exc).with_traceback(tb)
+                raise FakeSystemExit(*exc.args).with_traceback(tb)
```
omangin: Keep this one!

tests/test_usecase.py
omangin: Sorry, what we need is `FakeSystemExit, FakeSystemExit(*exc.args), exc_traceback=tb`.

tests/test_usecase.py
```diff
@@ -189,10 +189,23 @@ def tearDown(self):
 
 class TestAlone(CommandTestCase):
 
-    def test_alone_misses_command(self):
-        with self.assertRaises(FakeSystemExit):
```
omangin: We could also check that the error code is `3`.

tests/test_usecase.py
```diff
@@ -189,10 +189,23 @@ def tearDown(self):
...
+    def test_alone_prints_help(self):
```
omangin: We could keep this for python 3 only. (With a modification for the error message printed before the help.)

tests/test_usecase.py
omangin: Sorry, my message was written before yours.

You're probably right. At the end both versions are fine but let's go for the simplest. I guess any test suite would catch `BaseException` and not just `Exception` anyway.

wflynny:
Hey sorry for the little activity this past week.  Made these changes with a few tweaks.  Let me know what you think.

omangin:
Great job! Thanks a lot. I'm adding you to the contributors and merging…

Issue #90, opened 2017-09-27
Add the ability to list by years range


amartos:
Currently only papers published a specific year can be retrieved from the `pubs list year:YYYY` command. It would be nice to be able to give a range of years instead, for eg. *via* `pubs list year:YYYY-YYYY`.

omangin:
Yes! Supporting something like `year:2000-` would be great too.

PR #102, opened 2017-12-18
Feature: year ranges in query

omangin:
See #90.

Right now queries like `year:` and `year:-` would match any paper that has a year value which can be parsed as an int.

### DIFF
--- a/tests/test_queries.py
+++ b/tests/test_queries.py
@@ -3,6 +3,7 @@ import unittest
 import dotdot
 from pubs.commands.list_cmd import (AuthorFilter,
                                     FieldFilter,
+                                    YearFilter,
                                     _query_block_to_filter,
                                     get_paper_filter,
                                     InvalidQuery)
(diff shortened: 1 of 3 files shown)


wflynny:
I wonder if it makes more sense for the `*Filter` classes and maybe even the `get_paper_filter` function to be in a separate file (e.g. `filters.py`)?  The functionality implemented is more general than for what exists in `list_cmd.py`;  I can see their functionality being possibly useful in at least `export_cmd.py` and `remove_cmd.py` too (e.g. export every paper since 2017, or remove all papers by author X).

omangin:
Indeed, I was planning to move it when we effectively use the functionality  elsewhere, but you're right that there is no reason to wait.

omangin:
@wflynny, I moved the code as you suggested. Since you already had a look at it, could you make a quick github review before I merge it?

tests/test_queries.py
```diff
@@ -16,86 +14,113 @@
...
                                    metadata=fixtures.turing_metadata)
 
+
 class TestAuthorFilter(unittest.TestCase):
 
     def test_fails_if_no_author(self):
         no_doe = doe_paper.deepcopy()
         no_doe.bibentry['author'] = []
-        self.assertTrue(not _check_author_match(no_doe, 'whatever'))
+        self.assertTrue(not AuthorFilter('whatever')(no_doe))
```
wflynny: AuthorFilter returns `True`/`False`.  So shouldn't this be explicit with `self.assertIs(..., False)` or `self.assertFalse`?

tests/test_queries.py
```diff
@@ -16,86 +14,113 @@
...
+        self.assertFalse(AuthorFilter('doe', case_sensitive=True)(doe_paper))
+        self.assertTrue(AuthorFilter('dOe', case_sensitive=False)(doe_paper))
 
     def test_match_not_first_author(self):
-        self.assertTrue(_check_author_match(page_paper, 'motwani'))
+        self.assertTrue(AuthorFilter('motwani')(page_paper))
 
     def test_do_not_match_first_name(self):
-        self.assertTrue(not _check_author_match(page_paper, 'larry'))
+        self.assertTrue(not AuthorFilter('larry')(page_paper))
```
wflynny: Same comment regarding `assertFalse`/`assertIs(..., False)` instead of `assertTrue(not ...)`.  

Also, the test should probably be `lawrence` since `larry` isn't in the corresponding bibdata.

tests/test_queries.py
omangin: Sure! That's why reviews are useful…

omangin:
Thanks a lot @wflynny for the comments.

Issue #98, opened 2017-11-28
Alias plugin command descriptions


wflynny:
I've added quite a few custom commands via the alias plugin (about as many as the default number of commands).  It would be helpful if we could specify a brief description of these commands that would be available to `pubs --help`, for example.

I've modified the `Alias` class in `pubs/plugs/alias/alias.py` so that I can include a small description in the alias definition in the conf file.  For now that looks like the following example:

    [[alias]]
    fadd = !pubs_fancyadd.py  <- interactive version of `pubs add`

where I've separated the alias definition from it's description with `<-`.  This could obviously be changed.

Is there any interest in adding functionality like this out of the box?  And are there any specific requirements that should be met while implementing it?

omangin:
Thanks again! Having documentation of the alias function is indeed of interest! We have to check what makes the most sense for the configuration file syntax so that the parsing these does not interfere with the rest. Once that is fixed, it should be straightforward to implement. @benureau, any though?

omangin:
So, @wflynny. Sorry for only coming back to it now. It seems that if we want to keep the compatibility with the `configobj`, there are basically two options here.

The first one is to use a bash comment starting with `#`. The `shlex` parser would be used to get the content of the comment that would be set as the alias documentation. The issue of this approach is that the comment needs to be included within a string value otherwise it is removed when the config is parsed. This means that in the following only the second comment will be available to `pubs`:
```
fadd = !pubs_fancyadd.py  # interactive version of `pubs add`
fadd = "!pubs_fancyadd.py  # interactive version of `pubs add`"
```

The second option is a little bit more cumbersome to write but more compatible with `configobj`, more consistent, and probably more straightforward to implement. It would consist in having the possibility of setting the alias as part of a subsection as follows:
```
[[alias]]
# The first two ones have no description
evince = open --with evince
count = !pubs list -k "$@" | wc -l
[[[fadd]]]
# This alias include a description
command = !pubs_fancyadd.py
description = "interactive version of `pubs add`"
```

So, if you still want to implement it, please do. I'm personally more in favor of the second option. In the future it could also be used for other arguments of aliases (such as for autocompletion).

wflynny:
Yep that sounds easy enough.  Will probably get around to it over the weekend/holiday.

omangin:
Closed by #104.

PR #104, opened 2017-12-20
Feature: Alias descriptions in conf

wflynny:
See #98.

Allows aliases defined as currently supported and as subsections of the `[[alias]]` section of `conf.py` of the form:

    [[[alias-name]]]
    command = alias-command-definition
    description = alias-description

Note that aliases defined this way must be placed at the end of the `[[alias]]` section or they will gobble up aliases defined in the current format.  This is a limitation of ConfigObj sectioning.


omangin:
Great job @wflynny!

PR #105, opened 2018-01-06
[Fix #95] robust handling of DOIs

### LINKED ISSUE #95: robust handling of DOIs
Currently, only `pubs add -D 10.1007/s00422-012-0514-6` works. But given how DOIs are frequently displayed in webpages and articles, pubs should also accept:

```
pubs add -D https://doi.org/10.1007/s00422-012-0514-6
pubs add -D http://doi.org/10.1007/s00422-012-0514-6
pubs add -D doi.org/10.1007/s00422-012-0514-6
```

wflynny:
Added DOI regex function to utils.py which is called in add_cmd.py
upon specifying a new DOI.  DOI validation applies directly on
argument parsing by using a custom argparse.Action.

Two concerns:  

1.  One thing that I don't like is that I imported `argparse` into `add_cmd.py` in order to set up the custom `argparse.Action` to validate DOIs at parse time.  
2.  Does the DOI parsing need to happen anywhere else?  I thought about adding it to `apis.doi2bibtex` but that is only ever called in `add_cmd.py`.

### DIFF
--- a/pubs/commands/add_cmd.py
+++ b/pubs/commands/add_cmd.py
@@ -1,3 +1,4 @@
+import argparse
 from ..uis import get_ui
 from .. import bibstruct
 from .. import content
@@ -7,13 +8,21 @@ from .. import templates
 from .. import apis
 from .. import color
 from .. import pretty
+from .. import utils
 
+class ValidateDOI(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        doi = values
+        new_doi = utils.standardize_doi(doi)
+        if not new_doi:
+            raise ValueError("Not a valid doi: %s", doi)
+        setattr(namespace, self.dest, new_doi)
 
 def parser(subparsers, conf):
     parser = subparsers.add_parser('add', help='add a paper to the repository')
     parser.add_argument('bibfile', nargs='?', default=None,
                         help='bibtex file')
-    parser.add_argument('-D', '--doi', help='doi number to retrieve the bibtex entry, if it is not provided', default=None)
+    parser.add_argument('-D', '--doi', help='doi number to retrieve the bibtex entry, if it is not provided', default=None, action=ValidateDOI)
     parser.add_argument('-I', '--isbn', help='isbn number to retrieve the bibtex entry, if it is not provided', default=None)
     parser.add_argument('-d', '--docfile', help='pdf or ps file', default=None)
     parser.add_argument('-t', '--tags', help='tags associated to the paper, separated by commas',
--- a/pubs/commands/add_cmd.py
+++ b/pubs/commands/add_cmd.py
@@ -10,13 +10,13 @@ from .. import color
 from .. import pretty
 from .. import utils
 
+
 class ValidateDOI(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
         doi = values
         new_doi = utils.standardize_doi(doi)
-        if not new_doi:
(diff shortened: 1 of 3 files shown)


pubs/commands/add_cmd.py
```diff
@@ -7,13 +8,21 @@
 from .. import apis
 from .. import color
 from .. import pretty
+from .. import utils
```
omangin: PEP8: _Surround top-level function and class definitions with two blank lines._

pubs/commands/add_cmd.py
```diff
@@ -7,13 +8,21 @@
...
 
+class ValidateDOI(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        doi = values
+        new_doi = utils.standardize_doi(doi)
+        if not new_doi:
+            raise ValueError("Not a valid doi: %s", doi)
```
omangin: I would move this (raising the exception) to `standardize_doi`.

pubs/utils.py
```diff
@@ -44,3 +46,36 @@ def resolve_citekey_list(repo, citekeys, ui=None, exit_on_fail=True):
...
+        re.compile(r'(10\.1207/[\w\d]+\&\d+_\d+)')
+        )
+
+    for doi_regex in doi_regexes:
+        match = doi_regex.search(doi)
+        if match:
+            new_doi = match.group(0)
+            break
+    else:
+        new_doi = None
```
omangin: This should raise the `ValueError` exception you have in the other file.

pubs/utils.py
```diff
@@ -44,3 +46,36 @@ def resolve_citekey_list(repo, citekeys, ui=None, exit_on_fail=True):
...
+    https://www.doi.org/doi_handbook/2_Numbering.html
+    https://www.crossref.org/blog/dois-and-matching-regular-expressions/"""
+    """ :returns standardized doi """
+    doi_regexes = (
+        re.compile(r'(10\.\d{4,9}/[-._;()/:A-z0-9\>\<]+)'),
+        re.compile(r'(10.1002/[^\s]+)'),
+        re.compile(r'(10\.\d{4}/\d+-\d+X?(\d+)\d+<[\d\w]+:[\d\w]*>\d+.\d+.\w+;\d)'),
+        re.compile(r'(10\.1021/\w\w\d+\+)'),
+        re.compile(r'(10\.1207/[\w\d]+\&\d+_\d+)')
+        )
```
omangin: Can't you join all these regexp with `|` operators? That way we can also get rid of the for loop below.

pubs/utils.py
```diff
@@ -44,3 +46,36 @@ def resolve_citekey_list(repo, citekeys, ui=None, exit_on_fail=True):
...
+    -   http://doi.org/<doi>
+    -   https://dx.doi.org/<doi>
+    -   http://dx.doi.org/<doi>
+    -   dx.doi.org/<doi>
+    -   doi.org/<doi>
+    and attempts to verify doi adherence to DOI handbook standards and
+    crossref.org advice:
+    https://www.doi.org/doi_handbook/2_Numbering.html
+    https://www.crossref.org/blog/dois-and-matching-regular-expressions/"""
+    """ :returns standardized doi """
```
omangin: Shouldn't this be in the same multi-line string?

wflynny:
Thanks @omangin, great comments.  I think I got everything.  Let me know if I missed anything.

omangin:
Perfect. Thanks again @wflynny!

Issue #110, opened 2018-01-23
pubs is not readily available in Window  CMD after pip installation


sjystu:
Hello,

I was wondering if there would be any way you could upgrade pubs installation process  for Windows command line users.

In Linux, you can use "pubs" in command line as soon as you install it.
However, in Windows it does not seem to be added to PATH.

When I installed another Python package called "gscholar" using pip, it was automatically added to Windows PATH so that users can employ the alias in the CMD.

Please consider it.

Appreciatively,

wflynny:
I think this just requires a change from `scripts=` to `entry_points=` in `setup.py` [[1]].  Looks like it may be recommended practice too [[2]].  

This should be too hard in theory, but has `pubs` been tested at all on Windows?  @omangin a Windows CI build would need to be made, correct?  How much work it that?

[1]: http://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation
[2]: https://stackoverflow.com/a/28119736/1634191

omangin:
Hi @sjystu. We indeed do not support Windows and I have honestly no idea about how `pip` installs things on it. However the `pubs` script that is currently used has little chance to execute since it explicitly refers to `/usr/bin/env python` as an interpreter.

I have created a branch with @wflynny's suggestion on using `entry_points` instead. You may try it to check whether that works. Note that it's likely that the rest of the code is not fully functional. We would be happy to get feedback on whether this works on Windows. (Windows support is however unfortunately far from our priority.)

Regarding switching to `entry_points`, it might be better practice since we delay more tasks to pip. The issue right now is that I am not fully sure it does not break autocompletion, which I need to double check. I'll open a PR regarding that.

omangin:
Closing for now. Please re-open if you tried the latest version of the code with the _entry-point_ method and have more feedback.

benureau:
@sjystu: could you tell us your version of Windows, python, and how it was installed?
@omangin: since Python 3.3, `#!/usr/bin/env python` is supported on Windows. See https://docs.python.org/3.6/using/windows.html#shebang-lines . If we support pubs on Windows one day, we might do it for >=3.3 only.

sjystu:
Thank you so much for your kind response.

 

I used Python 3.5 on Windows 10 64 bit system.

FYI, I used Anaconda.

Please let me know if you need any other information from me.

 

Thanks.

omangin:
[...]

PR #112, opened 2018-01-23
Feat/entry point

omangin:
Needs to double check for auto-completion. (See also #110.)

### DIFF
--- a/pubs/pubs
+++ b/pubs/pubs
@@ -1,8 +1,0 @@
-#!/usr/bin/env python
-# -*- coding:utf-8 -*-
-# PYTHON_ARGCOMPLETE_OK
-
-from pubs import pubs_cmd
-
-
-pubs_cmd.execute()
--- a/pubs/pubs_cmd.py
+++ b/pubs/pubs_cmd.py
@@ -1,3 +1,5 @@
+# PYTHON_ARGCOMPLETE_OK
+
 import sys
 import argparse
 import collections
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,11 @@ setup(
               'pubs.templates',
               'pubs.plugs',
               'pubs.plugs.alias'],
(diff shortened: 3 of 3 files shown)


omangin:
OK, I added it to `pubs_cmd.py`. That should do the job. Can you double check that it work?

I also noticed that the global completion does not seem to work when the package is installed in `develop` mode. In which case on needs to execute the `eval "$(register-python-argcomplete pubs)"`. Might be related to [argcomplete/#112](https://github.com/kislyuk/argcomplete/issues/112) that was apparently fixed but not in the latest version. It's minor though, so I am just putting it there for the record.

omangin:
Great. I'll merge then. Do you have `eval "$(register-python-argcomplete pubs)"` somewhere in your path or are you using the global completion mode?

wflynny:
@omangin  Hmm in a brand new environment, you're right; it doesn't work in `develop` mode.  But I can't seem to find where in my normal test environment `eval "$(register-python-argcomplete pubs)"`.  For example, in both environments, this is all I have under `complete`.

```
(pubs-test)$ complete -p
complete -o bashdefault -o default -F _python_argcomplete_global -D
```

In my bash history, I do have `eval $"(register-python-argcomplete pubs)"`, but I'm not sure where to look for it's effects other than `complete`/`env`.  Using bash `version 4.4.12(1)-release (x86_64-apple-darwin16.3.0)`.

Should we open this up as a new bug?

omangin:
So if it sets the completion for `pubs`, it should show up in `complete | grep pubs`. Otherwise it is using the global mode that is active for you (as shown in the output you pasted above).

In any case, since it is working in `install` mode, this is not a pubs issue but eventually an upstream bug on `argcomplete`. However they seemed to have merged some things related to that that is not yet in the latest available version so it's probably already fixed.

Issue #103, opened 2017-12-19
List command: query entries containing unicode


wflynny:
Related to #15, #87 and maybe #28, #10.

I'm currently aliasing `pubs list "author:<author-name>"` as a primitive search command.  This works well for most cases but fails if the author's name contains unicode characters.  It would be nice to have some fuzzy matching or something similar.  For example,

    pubs list author:Grun
    pubs list author:Gr{\"u}n

fails to return any documents by author "Grün".  A work around is to just do the search using truncated query `pubs list author:Gr`, though that could lead to problems with larger repositories.

omangin:
Thanks! There are actually several issues here.

1. The comparison during the search should occur on normalized unicode strings (except maybe when some option is set that we could call `--strict` or `--no-normalize`).

2. In your example, the author name should be converted to unicode before any manipulation is performed, by calling bibtexparser's `convert_to_unicode` customization. However I'm not in favor of changing the default to always apply the customization because this store the entries as unicode which would come into the way of user's preferences on how to store bibdata. If I'm correct not all version of latex accept unicode and that would also remove the protection of capitals within `{}`. The customization should only be applied for comparison and print. (Currently list also displays `\"u` if that's what in the bibdata.)

3. The search query should probably be compared to an ascii equivalent of the field's value. Following the same scheme as for case sensitivity: a search query with only unicode characters is compared to the ascii equivalent of the field (through [unidecode](https://pypi.python.org/pypi/Unidecode)) and a search query with non-unicode characters is compared to the unicode. The corresponding options to force the behavior would be `--transliterate` and `--no-transliterate`.

Obviously implementing #15 would solve 1 and 2.

PR #116, opened 2018-02-03
Fixes unicode comparison

omangin:
This PR adresses the first two elements mentioned in [my comment](https://github.com/pubs/pubs/issues/103#issuecomment-354887360) from #103.

The PR relies on recent additions to `bibtexparser` and some of the test rely on [python-bibtexparser#187](https://github.com/sciunto-org/python-bibtexparser/pull/187) being merged.This is why the tests don't pass.

This should probably be merged after `bibtexparser 0.6.3` is released.

### DIFF
--- a/pubs/commands/list_cmd.py
+++ b/pubs/commands/list_cmd.py
@@ -15,6 +15,8 @@ def parser(subparsers, conf):
                         default=None, dest='case_sensitive')
     parser.add_argument('-I', '--force-case', action='store_true',
                         dest='case_sensitive')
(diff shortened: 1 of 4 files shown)


benureau:
@omangin. So it seems we need a newer version of bibtexparser to make that work. But there are no recent release. I left a message on the corresponding bibtexparser issue, FYI.

omangin:
So the new bibtexparser has been released. See #118 for the test that was not passing. I rebased this PR on top of it so that the test make sense.

Issue #122, opened 2018-02-21
common strings in parser by default


benureau:
The issue is described here https://github.com/sciunto-org/python-bibtexparser/issues/192, and displayed in action here: https://travis-ci.org/benureau/pubs/jobs/344167378
A fix is coming.

PR #123, opened 2018-02-21
Fix/common strings

benureau:
see #122

### DIFF
--- a/pubs/endecoder.py
+++ b/pubs/endecoder.py
@@ -113,15 +113,10 @@ class EnDecoder(object):
     def decode_bibdata(self, bibdata):
         """"""
         try:
-            try:
-                entries = bp.bparser.BibTexParser(
-                    bibdata, common_strings=True,
-                    customization=customizations,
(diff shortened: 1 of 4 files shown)


pubs/endecoder.py
```diff
@@ -115,11 +115,12 @@ def decode_bibdata(self, bibdata):
         try:
             try:
                 entries = bp.bparser.BibTexParser(
-                    bibdata, homogenize_fields=True,
-                    customization=customizations).get_entry_dict()
+                    bibdata, common_strings=True,
+                    customization=customizations,
+                    homogenize_fields=True).get_entry_dict()
             except TypeError:
```
omangin: According to a5a653becd761bf56a7ac38e309c29a5ea3307a3, this was targeting older versions of bibtexparser. Since we now require a newer version, we should probably remove the exception catching altogether.

benureau: Indeed. I am merging with this change included.

PR #130, opened 2018-03-20
Adds tag list completion in `pubs add -t `

omangin:
Quite basic.

The auto completion feature probably will probably need some testing at some point… although it is not completely sure how to do that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,43 @@
-os:
-  - linux
-language: python
-python:
-  - "2.7"
-  - "3.3"
-  - "3.4"
-  - "3.5"
-  - "3.6"
+# list of environments to test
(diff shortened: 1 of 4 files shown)


benureau:
Seems that we have problems with pyfakefs. Might be due to the recent 3.4.1 release. I am looking into it.

benureau:
So it seems that pyfakefs 3.3 is working. I opened #131 to look into the problem with version 3.4.1.
I also updated the travis config so that Python 2.7 on OSX is tested. 
@omangin: can you review my change and merge?

omangin:
BTW, what was the rational behind using explicit matrix includes rather than the regular os x py-version in travis for linux?

PR #139, opened 2018-05-09
Add a command to open urls

ksunden:
Hello pubs maintainers,

I just recently found this project and really enjoy using it.

I thought a useful command would be to open up the url field in a web browser.
This will help with things like filling out docs.

For the brave:
`pubs list -k --no-docs | xargs pubs url`
Note: This crashed Firefox on my machine the first time I ran it (though it did succeed later attempts).
I can, however recommend
 `pubs list -k --no-docs | head | xargs pubs url`
Which will limit the number of tabs open.

Thanks,
Kyle

### DIFF
--- a/pubs/commands/url_cmd.py
+++ b/pubs/commands/url_cmd.py
@@ -2,23 +2,22 @@ from __future__ import unicode_literals
 
 import webbrowser
 
-from .. import p3
 from .. import repo
 from ..uis import get_ui
-from ..utils import resolve_citekey, resolve_citekey_list
+from ..utils import resolve_citekey_list
 
 
 
 def parser(subparsers, conf):
     parser = subparsers.add_parser('url',
-                                   help="open the url in the browser")
+                                   help="open a paper's url in the default web browser")
     parser.add_argument("citekey", nargs = '*',
                         help="one or more citeKeys to open")
     return parser
 
 
 def command(conf, args):
-    """Open the url for a publication."""
+    """Open the url of one or several papers in a web browser."""
 
     ui = get_ui()
     rp = repo.Repository(conf)
@@ -27,6 +26,7 @@ def command(conf, args):
         try:
             paper = rp.pull_paper(key)
             url = paper.bibdata['url']
+            ui.info('opening url {}'.format(url))
             webbrowser.open(url)
 
         except KeyError as e:
--- a/pubs/commands/url_cmd.py
+++ b/pubs/commands/url_cmd.py
@@ -23,13 +23,13 @@ def command(conf, args):
     ui = get_ui()
     rp = repo.Repository(conf)
 
-    for key in resolve_citekey_list(rp, args.citekey, ui=ui, exit_on_fail=True):
+    for key in resolve_citekey_list(rp, args.citekey, ui=ui, exit_on_fail=False):
         try:
             paper = rp.pull_paper(key)
             url = paper.bibdata['url']
             webbrowser.open(url)
 
         except KeyError as e:
-            ui.error('{} has no url'.format(key))
+            ui.warning('{} has no url'.format(key))
 
     rp.close()
--- a/pubs/commands/url_cmd.py
+++ b/pubs/commands/url_cmd.py
@@ -1,0 +1,35 @@
+from __future__ import unicode_literals
+
+import webbrowser
+
+from .. import p3
+from .. import repo
+from ..uis import get_ui
+from ..utils import resolve_citekey, resolve_citekey_list
+
+
+
+def parser(subparsers, conf):
+    parser = subparsers.add_parser('url',
+                                   help="open the url in the browser")
+    parser.add_argument("citekey", nargs = '*',
+                        help="one or more citeKeys to open")
+    return parser
+
+
+def command(conf, args):
+    """Open the url for a publication."""
+
+    ui = get_ui()
+    rp = repo.Repository(conf)
+
+    for key in resolve_citekey_list(rp, args.citekey, ui=ui, exit_on_fail=True):
+        try:
+            paper = rp.pull_paper(key)
+            url = paper.bibdata['url']
+            webbrowser.open(url)
+
+        except KeyError as e:
+            ui.error(key, 'has no url')
+
+    rp.close()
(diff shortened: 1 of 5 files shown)


pubs/commands/url_cmd.py
```diff
@@ -0,0 +1,36 @@
...
+
+    for key in resolve_citekey_list(rp, args.citekey, ui=ui, exit_on_fail=True):
+        try:
+            paper = rp.pull_paper(key)
+            url = paper.bibdata['url']
+            webbrowser.open(url)
+
+        except KeyError as e:
+            ui.error('{} has no url'.format(key))
+            print(key)
```
omangin: It seems that the `print` is not needed.

Also, you need to exit with a non-zero code if the `url` key does not exist.

omangin:
Hi Ksunden. That's a great idea. Thanks for contributing,

I do not have time right now to review your PR but see my comment. Also it would be good to add one or two tests of the behavior.

Additionally we would like to look into whether using the `webbrowser` package is better than directly calling `xdg-open` with the url. Do you have any suggestion on that?

pubs/commands/url_cmd.py
ksunden: Since the command supports opening multiple URLs at once, I explicitly do not want to exit at except time.

Perhaps that should be a `warning` rather than an `error` at that point, however

ksunden:
`webbrowser` is nice as it is built in and cross platform. (While I use systems with `xdg-open` myself, not everyone does)
It is also consistent with the `websearch` command, which also opens a browser.


I agree that I would like a test, just not sure what such a test looks like as the desired result is a browser opening. May have to try out behavior without a graphical environment to know what the test should look like. I did look to see if there were existing tests for the `websearch` command as the closest comparable, but did not see one to base off of.

I may add a flag to simply print the url, rather than open them. That would at least be easier to test, as it is just reading stdout (and stderr).

benureau:
Hi @ksunden, thanks for the the contribution. For the test, you may want to create a small testcase where you add a reference without a url, and run the command to obtain a warning. This will at least ensure that breaking changes to the command boilerplate are detected.

I am not sure about the name of the command, `url`. Seems like I could expect the url to be displayed, rather than opened. What about `webpage`? Or `webopen`? @omangin, any ideas? 

On that note, I feel like a info text such as:
```
info: opening url http://dx.doi.org/10.1093/mind/lix.236.433
```
would be good, especially to make clear the intent, if anything goes wrong with the command. And, as the url will usually redirect to a publisher website, it helps to keep the original url somewhere displayed.

pubs/commands/url_cmd.py
```diff
@@ -0,0 +1,35 @@
+from __future__ import unicode_literals
+
+import webbrowser
+
+from .. import p3
+from .. import repo
+from ..uis import get_ui
+from ..utils import resolve_citekey, resolve_citekey_list
```
benureau: `p3` and `resolve_citekey` are not used here.

benureau:
By the way, you can find examples of tests in the https://github.com/pubs/pubs/blob/master/tests/test_usecase.py file

ksunden:
I'm open to changing the name, or putting flags in, if you feel that is appropriate.

Perhaps something like `pubs url citekey` just prints the url (with no formatting whatsoever, one per line, with warning for no url to stderr vi `ui.warning`) to stdout, then add a flag `-o/--open` for actually opening in the browser. (with printing to `ui.info`)

I've started writing tests, but hit a snag: using webbrowser within the fake filesystem used in the test suite raises a `FileNotFoundError: [Errno 2] No such file or directory: '/dev/null'
`
It appears that the `pyfakefs` does not provide /dev/null, and the webbrowser module by default sends stdout and stderr there to avoid clogging up the terminal session. 
I was able to use `self.fs.CreateFile` to make a file called /dev/null, which caused a different error later: `OSError: [Errno 9] Bad file descriptor: 5`


It also does *better* when I use a text only browser, though oddly setting $BROWSER to do so still results in things being opened in firefox (as well as w3m).

It also messes with the terminal output, resulting in newlines and user input not being printed properly.

In text-only mode, it opened w3m for all tests, but still had the issue with messing with the terminal session. It seemed to work, though the output was cleared, so I could not see what was printed. `$?` was 0, however, indicating success.

I committed my tests to a branch on my fork to keep them isolated until I can figure out what needs to be done there. https://github.com/ksunden/pubs/tree/tests

benureau:
@ksunden thanks for all your work! That's really great.

I'll had a quick look at the `/dev/null` issue, and I think you are right, pyfakefs is not provided. But maybe it should. If you look at those lines:
https://github.com/jmcgeheeiv/pyfakefs/blob/42ea6a09759e1c0a66d4266601353031db0e98bc/pyfakefs/fake_filesystem.py#L2878
, it add the standard streams, but I wonder if `/dev/null` could not be added through `os.devnull`? I would be probably be worth it to raise the issue to the pyfakefs maintainers.

I'll have a talk with @omangin by the end of the week, and we'll decide on the name. The flag option is interesting, and it keeps the name of the command as the one of the field, simplifying things for the user.

benureau:
Hum, looking more into this, I think the best way to go is to use `mock` for this (I don't know why I did not think of this earlier). See for instance:
http://www.datadependence.com/2016/03/learn-to-mock-in-python/
If you go this route, you'll probably need to install the `mock` package, so that the tests remains compatible with python 2.7, and add it to `tests/requirements.txt` so that Travis can install it.

ksunden:
Mock seems to do the trick.

Thus far I only check that the info line is printed as expected, but can look to adding the Mock specific assertions to be even more complete

tests/test_usecase.py
```diff
@@ -153,6 +154,7 @@ def setUp(self, nsec_stat=True):
         super(DataCommandTestCase, self).setUp(nsec_stat=nsec_stat)
         self.fs.add_real_directory(os.path.join(self.rootpath, 'data'), read_only=False)
         self.fs.add_real_directory(os.path.join(self.rootpath, 'bibexamples'), read_only=False)
+        self.fs.CreateFile('/dev/null')
```
benureau: Is this still necessary?

benureau:
@ksunden don't worry about adding assert around mock. You've done an amazing work. I'd like to have @omangin's input before merging, and we may add flags to the command later, but besides that, everything is good for me.

benureau:
@ksunden, everything's ok. I am merging this. Thank you for all your hard work. As we discussed, we are thinking about adding a flag to display the url/open the web browser, so the command api might change once more. I'll be sure to tell you if it does.

PR #142, opened 2018-06-10
dialog after editing paper

### LINKED ISSUE #136: dialog when renaming a citekey with `pubs edit`
When a user edit a citekey with `pubs edit`, a dialog should confirm the change when the editing is done.

benureau:
Fix #136.

### DIFF
--- a/pubs/repo.py
+++ b/pubs/repo.py
@@ -141,6 +141,13 @@ class Repository(object):
             pass
 
     def rename_paper(self, paper, new_citekey=None, old_citekey=None):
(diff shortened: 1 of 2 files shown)


pubs/repo.py
```diff
@@ -141,6 +141,13 @@ def pull_docpath(self, citekey):
             pass
 
     def rename_paper(self, paper, new_citekey=None, old_citekey=None):
```
omangin: Is the name of the method still good to reflect this behavior?

Issue #145, opened 2018-07-24
Add references with arXiv ID


joe-antognini:
It would be very convenient if papers could be added with their arXiv ID the same way that they can be added with their DOI since many arXiv papers don't have DOIs:

```
$ pubs add --arxiv math/0211159
```

or 

```
$ pubs add -X math/0211159
```

The way I would see this working is that if the arXiv paper already has a DOI (which can be obtained from the arXiv API), we would use the DOI API call.  If not (e.g., if the paper is unpublished), then we would generate a bibtex entry.

One downside of this feature is that if a paper is later published and gets a DOI, the user would have to manually add the updated information (DOI, journal, etc.) or else the reference information will be stale.  (Perhaps another feature would be a command like `pubs arxiv-update` or something that would check to see if the arXiv paper now has a DOI associated with it.)

I have a WIP that I can submit as a PR if the pubs team thinks this is a good idea.

benureau:
Hi @joe-antognini. Indeed, with the increase prevalence of arXiv, I have been thinking about doing that for quite a while. As an added bonus, with arXiv, we could arrange for the pdf to be downloaded as well. Do submit your work in progress, I think this could be a great addition to pubs.

benureau:
Closed by #146, #155.

PR #146, opened 2018-07-24
Arxiv support for pubs add.

joe-antognini:
This addresses issue #145 by adding arXiv support for the `pubs add` command.  Now a user can add a paper using its arXiv ID with

```
pubs add --arxiv 1234.1234
```

or

```
pubs add -X 1234.1234
```

If the arXiv paper has a DOI associated with it (according to the arXiv API), the usual DOI API call will be used to generate the bibtex.  If there is no DOI associated with it then it will generate a bibtex entry on its own.

### DIFF
--- a/pubs/apis.py
+++ b/pubs/apis.py
@@ -40,19 +40,20 @@ def arxiv2bibtex(arxiv_id):
     if 'title' not in entry:
         ui = get_ui()
         ui.error('malformed arXiv ID: {}'.format(arxiv_id))
-    if 'arxiv_doi' in entry:
-        return doi2bibtex(entry['arxiv_doi'])
+        bibtex = None
+    elif 'arxiv_doi' in entry:
+        bibtex = doi2bibtex(entry['arxiv_doi'])
     else:
         # Create a bibentry from the metadata.
-        bibtext = '@misc{{{},\n'.format(arxiv_id)
-        bibtext += 'Author = {'
+        bibtex = '@misc{{{},\n'.format(arxiv_id)
+        bibtex += 'Author = {'
         for i, author in enumerate(entry['authors']):
-            bibtext += author['name']
+            bibtex += author['name']
             if i < len(entry['authors']) - 1:
-                bibtext += ' and '
-        bibtext += '},\n'
-        bibtext += 'Title = {{{}}},\n'.format(entry['title'].strip('\n'))
-        bibtext += 'Year = {{{}}},\n'.format(entry['published_parsed'].tm_year)
-        bibtext += 'Eprint = {{arXiv:{}}},\n'.format(arxiv_id)
-        bibtext += '}'
-        return bibtext
+                bibtex += ' and '
+        bibtex += '},\n'
+        bibtex += 'Title = {{{}}},\n'.format(entry['title'].strip('\n'))
+        bibtex += 'Year = {{{}}},\n'.format(entry['published_parsed'].tm_year)
+        bibtex += 'Eprint = {{arXiv:{}}},\n'.format(arxiv_id)
+        bibtex += '}'
+    return bibtex
--- a/pubs/apis.py
+++ b/pubs/apis.py
@@ -1,7 +1,9 @@
 """Interface for Remote Bibliographic APIs"""
 
 import requests
+import feedparser
 from bs4 import BeautifulSoup
+from uis import get_ui
 
 
 def doi2bibtex(doi):
@@ -25,3 +27,32 @@ def isbn2bibtex(isbn):
     citation = soup.find("textarea").text
 
     return citation
+
+
+def arxiv2bibtex(arxiv_id):
+    """Return a bibtex string of metadata from an arXiv ID"""
+
+    url = 'https://export.arxiv.org/api/query?id_list=' + arxiv_id
+    r = requests.get(url)
+    feed = feedparser.parse(r.text)
+    entry = feed.entries[0]
+
+    if 'title' not in entry:
+        ui = get_ui()
+        ui.error('malformed arXiv ID: {}'.format(arxiv_id))
+    if 'arxiv_doi' in entry:
+        return doi2bibtex(entry['arxiv_doi'])
+    else:
+        # Create a bibentry from the metadata.
+        bibtext = '@misc{{{},\n'.format(arxiv_id)
+        bibtext += 'Author = {'
+        for i, author in enumerate(entry['authors']):
+            bibtext += author['name']
+            if i < len(entry['authors']) - 1:
+                bibtext += ' and '
+        bibtext += '},\n'
+        bibtext += 'Title = {{{}}},\n'.format(entry['title'].strip('\n'))
+        bibtext += 'Year = {{{}}},\n'.format(entry['published_parsed'].tm_year)
+        bibtext += 'Eprint = {{arXiv:{}}},\n'.format(arxiv_id)
+        bibtext += '}'
+        return bibtext
--- a/pubs/apis.py
+++ b/pubs/apis.py
@@ -11,6 +11,45 @@ class ReferenceNotFoundException(Exception):
     pass
 
 
+def get_bibentry_from_api(id_str, id_type, rp):
+    """Return a bibtex string from various ID methods.
+
+    This is a wrapper around functions that will return a bibtex string given
+    one of:
+
+    * DOI
+    * IBSN
+    * arXiv ID
+
+    Args:
+        id_str: A string with the ID.
+        id_type: Name of the ID type.  Must be one of `doi`, `isbn`, or `arxiv`.
+        rp: A `Repository` object.
+
+    Returns:
+        A bibtex string.
+
+    Raises:
+        ValueError: if `id_type` is not one of `doi`, `isbn`, or `arxiv`.
+    """
+
+    id_fns = {
+        'doi': doi2bibtex,
+        'isbn': isbn2bibtex,
+        'arxiv': arxiv2bibtex,
+    }
+
+    if id_type not in id_fns.keys():
+        raise ValueError('id_type must be one of `doi`, `isbn`, or `arxiv`.')
+
+    bibentry_raw = id_fns[id_type](id_str)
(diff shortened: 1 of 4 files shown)


pubs/apis.py
```diff
@@ -25,3 +27,33 @@ def isbn2bibtex(isbn):
...
+    """Return a bibtex string of metadata from an arXiv ID"""
+
+    url = 'https://export.arxiv.org/api/query?id_list=' + arxiv_id
+    r = requests.get(url)
+    feed = feedparser.parse(r.text)
+    entry = feed.entries[0]
+
+    if 'title' not in entry:
+        ui = get_ui()
+        ui.error('malformed arXiv ID: {}'.format(arxiv_id))
```
omangin: In general, we try to keep the calls to UI errors outside of the helpers functions and mostly at the command layer (for instance this might not be an error but only a warning, which can only be decided at the command level). 

Here I would suggest raising a dedicated exception. You may reuse the same exception for various API issues and set the correct error message here, then call something like `ui.error(ustr(e))` when the exception is caught in the command code.

pubs/apis.py
```diff
@@ -25,3 +27,33 @@ def isbn2bibtex(isbn):
...
+        bibtex = None
+    elif 'arxiv_doi' in entry:
+        bibtex = doi2bibtex(entry['arxiv_doi'])
+    else:
+        # Create a bibentry from the metadata.
```
omangin: Here you can use `bwriter` from the `bibtexparser` package (see https://github.com/pubs/pubs/blob/master/pubs/endecoder.py#L85); this would more likely yield a well-formatted bibtex (be more robust to escaping characters, etc.)

pubs/commands/add_cmd.py
```diff
@@ -86,21 +87,26 @@ def command(conf, args):
         if args.doi is None and args.isbn is None:
             bibentry = bibentry_from_editor(conf, ui, rp)
         else:
+            bibentry = None
```
omangin: Here the new logic is to try all submitted information in order of DOI, ISBN, Arxiv (eventually overwiting previous results) and then failing if none yield a well formatted result.

With regard to my comment above about using exceptions in the API code, it means that these exceptions should be caught here and yield warnings, not error (there may be a warning about the arxiv not being well formatted although a bibtex could be found through the provided DOI or ISBN). At the end only, an error is printed and the `ui.exit` called if none of the API provided a result.

Currently it could occur that the DOI yield a bibtex that would be overwritten by the Arxiv API returning `None` in the case of https://github.com/pubs/pubs/pull/146/files#diff-a517beca1cb0eb07bc6a2a9ae7e7f544R40. Using an exception as explained above would avoid this and lead to an overall more robust behavior.

We will probably need a [_usecase_](https://github.com/pubs/pubs/blob/master/tests/test_usecase.py) to check this.

pubs/apis.py
joe-antognini: This makes sense.

pubs/apis.py
joe-antognini: Good idea.  I've now constructed a `BibDatabase` and then returned the result from running `bibtexparser.dumps` on it.

pubs/commands/add_cmd.py
joe-antognini: Yes, good catch.  I've changed it now so that if one of these fails, the program will exit.  I.e., if an invalid DOI is presented along with a valid arXiv ID, the program will still fail.  (Although, if a valid DOI is presented along with an invalid arXiv ID nothing will happen.)  I think this is the least surprising behavior since the user will probably want to know that they have provided invalid information to the program rather than have the program decide what is correct.  But I could be persuaded otherwise.  

I agree that we should have a usecase test to check this, though I haven't added it yet.

omangin: > I think this is the least surprising behavior since the user will probably want to know that they have provided invalid information

The thing is, right now, the program may fail when a DOI/arXiv/ISBN source reference is valid but not found.

More generally, I see two consistent behaviors we may consider:
 1. accept several of these options (`-X`, `-S`, `-I`) to be set and use one that works (with some precedence rules and show warnings for the others), or open an editor,
 2. accept at most one of these to be set and fail when it is not found.

In particular I don't think it is consistent to require references to be found but then eventually silently ignore the result of the arXiv source because a DOI was set too.

In my opinion option 2. has the least potential ambiguity for the user since it is clear that if a DOI is set it will be used or the program will fail. Also in 1. one could expect that the resulting data (e.g. from a DOI and arXiv) are merged which is not the case. Also, in most cases only one of these options will be used.

pubs/commands/add_cmd.py
```diff
@@ -84,28 +84,33 @@ def command(conf, args):
...
-                    ui.error('invalid isbn {} or unable to retrieve bibfile from it.'.format(args.isbn))
-                # TODO distinguish between cases, offer to open the error page in a webbrowser.
-                # TODO offer to confirm/change citekey
-            if args.arxiv is not None:
-                bibentry_raw = apis.arxiv2bibtex(args.arxiv)
-                bibentry = rp.databroker.verify(bibentry_raw)
-                if bibentry is None:
-                    ui.error('invalid arxiv id {} or unable to retrieve bibfile from it.'.format(args.arxiv_id))
-            if bibentry is None:
+            try:
```
omangin: Here there is a clear pattern for refactoring! (e.g. move the duplicated code outside of the `if`s)

omangin:
@joe-antognini, good fixes. See my comments for the last improvements.

In particular, regarding [whether we should prevent several of the arXiv, DOI, and ISBN options to be set simultaneously](https://github.com/pubs/pubs/pull/146/commits/ccdbe72eb7bffa11c8cc8e9d5377ec519ce43486#r205967184).

benureau:
What about prompting the user with a dialog if a doi is found for a arXiv paper and let him choose?

omangin:
> What about prompting the user with a dialog if a doi is found for a arXiv paper and let him choose?

Do you mean from a call of the form `pubs add -X ...` or `pubs add -D ... -X ...`? There is indeed a case in which the user sets an arXiv id but a more up-to-date bibtex is available through the DOI that arXiv returns (e.g. if the paper was later published and the authors manually set the external DOI in arXiv; however I am not sure it would be the DOI returned by arXiv, which we need to check). In that case a dialog would be meaningful but this would be the matter of another issue.

In the case where several sources are set, I still believe trying to handle the variety of situation automatically (or through dialogue) is difficult and might 1. induce a lot of complexity in the code, 2. make the behavior of `pubs` hard to predict. What I envision for this PR is a behavior like:
```bash
> pubs add -D ... # the user choses to try the DOI first
error: invalid doi ... or unable to retrieve bibfile from it.
> pubs add -X ... # the user then moves to the arXiv id
# if one was to try:
> pubs add -X ... -D ...
error: it is not allowed set several of `--arxiv`, `--doi`, and `--isbn` arguments together
```

We can move to handling more complex cases later but I would make it a separate issue because in my opinion this needs to get done right, which can take a lot more time. I am not in favor of releasing an half-baked version of the more complex behavior.

benureau:
@joe-antognini: We are nearing release of a new version, and would really like to have your arXiv support included in it. If you want, I can implement the last "DOI or arXiv, choose one" mechanism, so that we can merge your contribution to master.

joe-antognini:
I won't be able to get to the "DOI or arXiv, choose one" mechanism until the weekend so if you want to implement it before then, please do!

benureau:
Ok, so I am working on it right now. I looked at the `arxiv2bibtex()` code and I am wondering if we could rather fetch the bibtex from the NASA ADS service that seems associated with every article in arXiv. We would basically do:
```python
import requests
arxiv_id = '1804.08730'
url = "http://adsabs.harvard.edu/cgi-bin/bib_query?arXiv:{}&data_type=BIBTEX".format(arxiv_id)
bibtex = requests.get(url).text
```

But I don't know if this is a robust approach. I looked more into Bibtex and arXiv, and there are new-style references, and old-style references (before 2007), and doing that above would remove the need to care about that.

@omangin, @joe-antognini: thoughts?

omangin:
I don't know much about the NASA ADS query system but I would say that the current implementation
1. exists,
2. uses the official arXiv API.

So, if we don't have further guarantees that the ADS is covering all arXiv papers and not just astrophysics and that it is going to be stable over time, I would postpone this discussion to another issue.

joe-antognini:
I was thinking about leveraging ADS, but I don't think they get very much information for non-astronomy papers.  Here's a recent paper, for instance, that has a DOI associated with it, but ADS doesn't know about the DOI: https://arxiv.org/abs/1807.10906.  My suggestion would be to add a feature down the line that lets you add a reference by the ADS bibcode.

benureau:
@joe-antognini: Good find; indeed, that makes ADS insufficient. Ok, we use the arXiv API, it seems the more precise and robust choice. I am going to make the retrieved reference an `@article`, and add eprint fields according to section 3.13.7 of the [Biblatex manual](http://mirrors.ibiblio.org/CTAN/macros/latex/contrib/biblatex/doc/biblatex.pdf).

joe-antognini:
The latest push now makes DOI, ISBN, and arXiv IDs mutually exclusive arguments.  I think this makes sense because a user could accidentally add incompatible IDs and it's better for us just to fail in this case and require them to choose one or the other.

omangin:
@benureau What is the purpose of the [`requirements.txt`](https://github.com/pubs/pubs/commit/c513870132999bf1a5ffdeb916ba9b4e6636d8ba)? Is it only for documentation? It does not seem to be used anywhere and is redundant with the `setup.py`. Could we also replace the `test/requirements.txt` by a `tests_require` within the `setup.py`?

benureau:
@joe-antognini: nicely done.

I did some work to handle any error arxiv might throw at us, as well as handling old/new style references, I am merging this pull request into a new branch and continuing the work there.

@omangin: hum... not really. It's nice to have both. I'll make it specialized for dev work. Yes, `test_requires` is worthwhile, but only if we configure the rest so that `python setup.py test` works.

PR #149, opened 2018-07-25
Fix #87

### LINKED ISSUE #87: More robust pubs list
`list` command code needs to be more robust in case one or more bibdata entry are unreadable. I had one bib file empty, and I got
```
$ pubs list
error: could not parse provided bibdata:

$ 
```

omangin:
Also includes updates of editor dialog when a Bibtex decoding error occurs. See commit messages.

### DIFF
--- a/pubs/commands/import_cmd.py
+++ b/pubs/commands/import_cmd.py
@@ -13,6 +13,10 @@ from ..uis import get_ui
 from ..content import system_path, read_text_file
 
 
+_ABORT_USE_IGNORE_MSG = "Aborting import. Use --ignore-malformed to ignore."
+_IGNORING_MSG = " Ignoring."
+
+
 def parser(subparsers, conf):
     parser = subparsers.add_parser('import',
             help='import paper(s) to the repository')
@@ -24,10 +28,12 @@ def parser(subparsers, conf):
             help="one or several keys to import from the file")
     parser.add_argument('-O', '--overwrite', action='store_true', default=False,
             help="Overwrite keys already in the database")
+    parser.add_argument('-i', '--ignore-malformed', action='store_true', default=False,
+            help="Ignore malformed and unreadable files and entries")
     return parser
 
 
-def many_from_path(ui, bibpath):
+def many_from_path(ui, bibpath, ignore=False):
     """Extract list of papers found in bibliographic files in path.
 
     The behavior is to:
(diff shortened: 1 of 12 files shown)


pubs/commands/import_cmd.py
```diff
@@ -49,11 +49,17 @@ def many_from_path(bibpath):
...
+            biblist.append(coder.decode_bibdata(read_text_file(filepath)))
+        except coder.BibDecodingError:
+            ui.error("Could not parse bibtex at {}. Aborting import.".format(filepath))
+            ui.exit()
 
     papers = {}
     for b in biblist:
         for k, b in b.items():
+            if k in papers:
+                ui.warning('Duplicated citekey {}. Keeping last.'.format(k))
```
benureau: "Keeping last" -> "Keeping the last one".

pubs/commands/import_cmd.py
```diff
@@ -49,11 +49,17 @@ def many_from_path(bibpath):
 
     biblist = []
     for filepath in all_files:
-        biblist.append(coder.decode_bibdata(read_text_file(filepath)))
+        try:
+            biblist.append(coder.decode_bibdata(read_text_file(filepath)))
+        except coder.BibDecodingError:
+            ui.error("Could not parse bibtex at {}. Aborting import.".format(filepath))
+            ui.exit()
```
benureau: Do we want to abort here? I often found that when there is a problem, I want as much data about it as possible. Even if we decide here to abort the import, we could parse the other files and provide feedback for them too. Are all files unreadable? Just one? Just a bunch? Imagine there are 10 files that have the same problem. I would like to know their name right away, rather than fixing one, doing the import again, and getting a new name, 10 times.

tests/test_usecase.py
```diff
@@ -682,16 +691,40 @@ def test_conf(self):
...
-                    ('pubs edit Page99', ['', 'a']),
+                    ('pubs edit Page99', ['', 'n']),
                    ]
             self.execute_cmds(cmds)
 
+    def test_editor_succeeds_on_second_edit(self):
+        cmds = ['pubs init',
+                'pubs add data/pagerank.bib',
+                ('pubs edit Page99', [
+                    '', 'y',
```
benureau: You could stress it more than the empty string, I think. What about:
`test_editor_succeeds_on_third_edit` with the typical missing colon: `@misc{Page99, title="TTT" author="X. YY"}`?

pubs/commands/import_cmd.py
omangin: True but my point was that if there is an issue, the user might want to fix the files before importing. In particular it might be an issue to have to re-run an import on the same bunch of partly-imported files because of having to force the overwrite option on the second run.

Since it seems that both are valid use cases, I will add an option for ignoring read errors (whether the file is not found or the bibtex is malformed) and mention it in error messages. That way it will also make the behavior with wrongly formatted files coherent with entries that do not pass the check.

pubs/commands/import_cmd.py
```diff
@@ -13,6 +13,10 @@
 from ..content import system_path, read_text_file
 
 
+_ABORT_USE_IGNORE_MSG = "Aborting import. Use --ignore-malformed to ignore."
+_IGNORING_MSG = " Ignoring."
```
benureau: Ignore is a very transitive verb.

omangin: …

PR #159, opened 2018-08-09
Fix/doc add

### LINKED ISSUE #144: "add" command ignores doc_add=move
Steps to reproduce:

1. Set `doc_add = move` in `.pubsrc`
2. Run `pubs add -D foo -d some.pdf`

**Expected behavior**: `some.pdf` is moved to the document directory.

**Actual result**: `some.pdf` is copied to the document directory.

omangin:
Fixes #144: behavior of add_copy mode during add.

- correctly handles add_copy mode and configuration,
- add option top force 'copy' mode (since the default had been changed in between from 'copy' to 'move' this is now needed),
- fixes assumption in one test that the default is 'copy' (in other words the test was broken and ensuring that the functionality was broken too.),
- do not try to delete the source when it is an URL.


pubs/commands/import_cmd.py
```diff
@@ -8,28 +8,32 @@
...
+    parser.add_argument(
+        'keys', nargs='*',
+        help="one or several keys to import from the file")
+    parser.add_argument(
+        '-O', '--overwrite', action='store_true', default=False,
+        help="Overwrite keys already in the database")
+    parser.add_argument(
+        '-i', '--ignore-malformed', action='store_true', default=False,
+        help="Ignore malformed and unreadable files and entries")
+    add_doc_add_arguments(parser, move=False)
```
benureau: What the rational for not allowing move here? A comment in the code could be helpful.

omangin: Well,
– the not-so-convincing reason was to match the current behavior,
– the real reason was that I was not confortable having a default behavior that moves files because of some mention in the bibtex that is not clearly visible from the user. More precisely having some `file = "..."` in the imported bibtex would mean having these files moved around. Since this is not as clear as in the `add` command in which the user explicitly passes a `--doc` argument.

I believe nonetheless that it would be good to have a move command line argument for the `import` command but it should probably not share the config option with the `add` command and its default. (This somehow makes sense since the config option is called `doc_add`.) We could eventually add a config option called `doc_import` to make it event clearer.

As a matter of fact, I am also not sure that `doc_add = 'move'` is the right default and I would rather have a default to `copy`.

Any opinion on that?

benureau: I see. I don't believe that a `doc_import` is needed. We don't import regularly. We should just support a command line argument down the line.
I think `copy` as a default is a good idea.

benureau:
Perfect.

Issue #166, opened 2018-08-26
description in PyPi


benureau:
Currently, the description is empty in PyPi (https://pypi.org/project/pubs/). We should make sure it is not for the next release.

benureau:
closed by 8fe24140.

Issue #165, opened 2018-08-26
[regression] error: 'module' object has no attribute 'ReferenceNotFoundException'


yurivict:
Version ```0.8.0``` fails:

```
$ pubs add -d ~/Downloads/x.pdf -D {some-DOI}
error: 'module' object has no attribute 'ReferenceNotFoundException'
```

```pubs-0.7.0.104``` didn't have this problem.

**Comment by benureau** [comment]
Thanks for this. I forgot to properly test a function. I'll push a fix ASAP.

**Comment by benureau** [comment]
@yurivict: I pushed a fix in the `fix/doi_stats` branch. It will be merged soon into master and we'll publish a hotfix release (0.8.1) right after that. In the meantime, you can install the updated version with:
```
pip install -U git+https://github.com/pubs/pubs@fix/doi_stats
```

yurivict:
Thank you!

benureau:
Included in version 0.8.1. Closed by 8fe24140.

yurivict:
Thank you for fixing it!
I updated the port.

Issue #164, opened 2018-08-26
Empty pubs repository: 0 division error with statistics


wschuell:
If the repository is empty, pubs statistics raises a zero division error

```
➜  ~ pubs statistics      
error: float division by zero
Repository statistics:
➜  ~
```

benureau:
Thanks William! That's obviously something we did not check for. Will be fixed very soon.

benureau:
Ok it's fixed. The patch should be merged in master soon. Henceforth, you'll get:
```
$ pubs statistics 
Your pubs repository is empty.
```
I'll have to think about a witty remark to add to that.

wschuell:
Or maybe add useful info like:
```
Your pubs repository is empty.
You can add or import new entries using 'pubs add' or 'pubs import'.
If you want to synchronize your repository between different computers, simply synchronize the '$HOME/.pubs' repository. But you will still need to execute 'pubs init' once on each computer.
```
The last sentence could be removed if during install (or at first run if not found) the .pubsrc is automatically created; but that would be another issue.

benureau:
Hum. A link towards the first steps section of the documentation could do that. But let's be real: only very specific persons will use `pubs statistics` before `pubs add`. Congratulation William, you made the cut! :)

benureau:
Oh, and you can place the `.pubs` directory anywhere you want, for instance in your <sync provider of you choice> folder. Or even move it after the fact, if you edit your `.pubsrc` appropriately.

benureau:
Included in version 0.8.1. Closed by 8fe24140.

Issue #163, opened 2018-08-23
`pubs` should not be an error


benureau:
```
$ pubs
error: Too few arguments!

usage: pubs [-h] [--version]
            {init,conf,add,rename,remove,list,edit,tag,statistics,doc,note,export,import,websearch,url,open}
            ...
```
should be:
```
$ pubs
usage: pubs [-h] [--version]
            {init,conf,add,rename,remove,list,edit,tag,statistics,doc,note,export,import,websearch,url,open}
            ...
```
Also, and I have not check if it's present anywhere else, but I feel that the exclamation mark is overly aggressive and unnecessary.

benureau:
Yes, and why I  would be inclined to consider `pubs` as an alias of `pubs --help`, what's crucial to me is to remove the error message.

benureau:
Included in version 0.8.1. Closed by 8fe24140.

PR #167, opened 2018-08-27
Fix/doi stats

benureau:
Fix for #163, #164 and #165, and provides an opportunity to address #166. 

@omangin: could you review, (and merge and release 0.8.1 if you have time). #165 is particularly problematic (entirely my fault). I still need to add more usecase tests to guard against future issues like #165, but that can wait for 0.9.0.

### DIFF
--- a/pubs/apis.py
+++ b/pubs/apis.py
@@ -8,6 +8,8 @@ from bibtexparser.bibdatabase import BibDatabase
 import feedparser
 from bs4 import BeautifulSoup
 
+from . import endecoder
+
 
 class ReferenceNotFoundError(Exception):
     pass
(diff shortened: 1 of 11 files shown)


omangin:
Great job! Nothing to change.

I still think we should stick to reviewing all PRs, though. It's true that it is frustrating for straightforward fixes like this and can induce delays but maybe we can find a better way to address that than just skipping the reviews. Or maybe we do need exceptions but then we probably should make clear what they are. Here it would be _"straightforward fixes by one of the maintainers"_ but this sounds like a slippery slope to me. Any though?

benureau:
I did not skip the review because the fix was straightforward. I skipped it because it was an horrible bug (#165) in the recently published release affecting all users... 'gave you 24 hours before I went ahead though. I think this is a fine heuristics: "outstanding bug in released code, without quick response (24 hours) from any reviewer". With the implied responsibility to implement any changes requested from a reviewer, even after merging. What do you think?

benureau:
To be fair, I could have sticked to fixing #165. Indeed, the rest was slippery.

omangin:
Fair enough. I just wanted to get your thoughts on this.

Issue #45, opened 2015-12-16
list should indicate an attached document


73:
`pubs list` should indicate that a document it attached to a publication. This could be a special symbol like `+` or something....  
--73

humm:
Yes. I do agree. In the meantime, you have the `--no-docs` option to hunt which docs are without a document.

73:
Yeah. But from a listing I would like to see on which key I may call pubs
doc open on.
--73

omangin:
One idea to implement this would be to add *special tags* which values would be defined by pubs. The way gmail implements *inbox*, *sent*, *spam*.

The pros are that the code already exists for display and search. So even if the functionality is not reused it might already be beneficial with respect to the quantity of code to write.

On the other side, as it is always touchy to introduce reserved names in a namespace, we have to check whether it worth it, which in my opinion translates to *whether the feature will be used for something else*. Anyway, to limit the impact on the namespace I would suggest using a prefix for this kind of tags.

Re-use candidates are:
- a tag to know whether the document has a note.
- a tag to know the format of the document too (pdf, djvu, ps).
- any tag a plugin would like to offer (for example a "recent" tag, a "CitedBy+ReferenceToMyPubli" tag, ...)

Obviously these ideas are quite remote and not so useful, so the functionality is not yet required. However if we want to reserve the namespace it is better to do it know. Or maybe you have other ideas.

73:
I thought about an autogenerated tag like `+`which cannot be given. It would (or should) cause an syntax error. But I would not safe it in the sense of only showing it if `paper.docpath is not None`. 
Maybe there should ne `tags`(given by the user) and `labels`given by pubs?
This issue was about a view enhancement (no data to be stored). I see good points in your idea, but it is not necessary for this issue and therefore maybe postponed for another issue (with another usecase)?      
--73

benureau:
Ok, #168 closes this.

PR #168, opened 2018-08-28
doc extension visible in pubs list

benureau:
A quick fix for #45.
@omangin: thoughts? The idea is provide a simple solution, I deliberately avoided any sort of smart tag as discussed in the issue.

### DIFF
--- a/pubs/pretty.py
+++ b/pubs/pretty.py
@@ -67,8 +67,11 @@ def paper_oneliner(p, citekey_only=False):
         bibdesc = bib_oneliner(p.bibdata)
         doc_str = ''
         if p.docpath is not None:
-            doc_str = color.dye_out(' [{}]'.format(os.path.splitext(p.docpath)[1][1:]),
-                                    'tag')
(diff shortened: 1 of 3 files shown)


pubs/pretty.py
```diff
@@ -64,8 +65,12 @@ def paper_oneliner(p, citekey_only=False):
         return p.citekey
     else:
         bibdesc = bib_oneliner(p.bibdata)
+        doc_str = ''
+        if p.docpath is not None:
+            doc_str = color.dye_out(' [{}]'.format(os.path.splitext(p.docpath)[1][1:]),
```
omangin: Currently it is possible to add a document without an extension. I am not sure whether we should allow it in the first place or not but this would raise an exception here because of the `[1:]` indexing. You could just replace it with `'NOEXT'` or something here when the extension is empty.

omangin:
@benureau Please merge if you are OK with the addition.

benureau:
Ok, nice. Thanks! Merging

PR #174, opened 2018-09-27
Fix 173 and release as 0.8.2

omangin:
- [X] fix readme syntax
- [x] fix python2 pip install
- [x] add to changelog and create new release
- [ ] add periodic Travis installation check


benureau:
@omangin: So it seems we have problems with pyfakefs on 2.7. I am looking into those. Do you want me to review the PR now or is it still WIP?

omangin:
I guess it works either way since the issue seems independent to this PR. So it does not really matter for the review. For merging, what would be ideal is to fix the issue first, we can then rebase this and release the new version with the fix.

benureau:
Ok, so the checks pass, except for Python 3.3. I did not look to much into it, seems an installation problem (update of setuptools?). Python 3.3 is not supported since the 29th of September, 2017, so I think we can drop support as well rather than spend time on this. I added an allow fail to 3.3 in the travis matrix.

omangin:
Travis can do periodic builds: https://docs.travis-ci.com/user/cron-jobs/.

These are however the same as the regular builds from the source code repository. Here I was thinking on a build that just calls `pip install pubs` and then checks that it works by running a few commands.

The only way I see to do that is to look at the `TRAVIS_EVENT_TYPE` variable and run a completely different sequence of actions then but I am afraid that this messes up the travis matrix.

benureau:
Why don't you had a `before_install` line `pip install pubs` in some of the python versions? This should not interfere with the tests. You can even do a `pubs --help` and then `pip uninstall pubs` right after.

benureau:
Ok, did just that in #179.

Issue #92, opened 2017-09-28
Add search through attached docs


amartos:
The `list` or `doc` commands could also be used to search throughout the attached docs for (combinations of) keywords. For now, I’m using the `pdfgrep` command to do that.

omangin:
That's a good idea too, although not a priority for now. I don't know how good is the result returned by `pdftotext` and how fast it is but this would probably require storing the text of the pdf in the cached metadata. Probably needs to be addressed together with #15.

hpdeifel:
Another easy option is [pdfgrep](https://pdfgrep.org), which already does caching and is often more efficient than `pdftotext`. E.g I have this alias:

```
[[alias]]
search = !pdfgrep --cache -rni "$@" ~/.pubs/doc
```

Which gives me case-insensitive full-text search with caching and prints the matching page numbers and lines.

omangin:
Very good point. The question is now whether this deserves a dedicated pubs command (or plugin) or should stay an alias. I see two arguments in favor of the former:
– your alias has to hard-code the document directory and there is currently no way to ask pubs for this information from the command line,
– an important feature of such a search command would be to return the citekeys of the found documents (the way `pubs list -k` does), which is also not trivial to yield through such an alias.

PR #180, opened 2018-12-17
Fixed argument passing for alias plugin

zamlz:
_(Disregard this if it wasn't intended)_
I was going through the alias code and realized that shell command doesn't use its arguments passed to it via the `.format()` function.

### DIFF
--- a/pubs/plugs/alias/alias.py
+++ b/pubs/plugs/alias/alias.py
@@ -55,7 +55,7 @@ class ShellAlias(Alias):
         as shell arguments.
         """
         subprocess.call(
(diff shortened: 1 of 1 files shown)


benureau:
@zamlz: thank you, that looks like a good fix. I'll probably implement a test for checking that this does not happen in the future before merging it.

benureau:
@zamlz: Hum, I was too quick before. This is intended. If you want to use the arguments passed to the code, add them to the alias code. This allow flexible code. See this example: https://github.com/pubs/pubs/issues/92#issuecomment-399111988

Closing this now. Don't hesitate to reopen if needed.

zamlz:
Oh I see, that makes sense.

Issue #172, opened 2018-09-26
'pubs conf' doesn't work with -c flag


KKPMW:
Just started using pubs so please forgive if I missed something.

Moved pubsrc file to another place and made an alias for pubs to start with -c flag.
Everything seems to work fine except `pubs conf` which still looks for configuration file in the old place:

> error: configuration file not found at /home/user/.pubsrc

benureau:
@KKPMW: Thanks for the report, and welcome to pubs. 

The mistake is ours. It boils down to not using these lines in the `pubs conf` code itself:
https://github.com/pubs/pubs/blob/8fe241404000fcad6321666e57a8cd1290d54b60/pubs/pubs_cmd.py#L51-L54

I'll fix it asap. In the meantime, the workaround is to edit the configuration file directly. This will work exactly the same.

PR #182, opened 2019-01-04
handle / in citekeys

### LINKED ISSUE #179: Unable to add this paper from arxiv. Any ideas?
For some reason I can't add [this particular arxiv paper](https://arxiv.org/abs/1803.10122). Other papers work just fine.
```
error: File does not exist: /home/zamlz/.pubs/bib/https:/doi.org/10.5281
```
Wondering if you guys knew what this meant.

benureau:
So, two embarrassing things. Pubs was unable to handle any citekey with a `/` in it, which included all old-style arXiv papers. Which meant that the example given in the readme `pubs add -X math/9501234 -d article.pdf` never worked. This fixes #179 (which was twice problematic). The solution is to generate a LastnameYear citekey whenever `/` is found in the bibentry citekey.

And the second thing was that readme's `pubs add Loeb2012_downloaded.pdf Loeb_2012` was incorrect and should have been `pubs doc add Loeb2012_downloaded.pdf Loeb_2012`. I added a test that tries to test the readme as close as possible.

If you spot the code for an unknown "bibtex" command, see #183.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,6 +63,7 @@ matrix:
       python: 2.7
       env:
         - TO_TEST=INSTALL
+      if: type = cron
     - os: linux
       language: python
       dist: xenial
@@ -70,16 +71,19 @@ matrix:
       python: 3.7
       env:
         - TO_TEST=INSTALL
+      if: type = cron
     - os: osx
       language: generic
       python: 2.7
       env:
         - TO_TEST=INSTALL
+      if: type = cron
     - os: osx
       language: generic
       python: ">=3.6"
       env:
         - TO_TEST=INSTALL
+      if: type = cron
 
   allow_failures:
     - python: 3.3


benureau:
Just added a fix for #172.

pubs/commands/__init__.py
```diff
@@ -18,3 +18,4 @@
 # bonus
 from . import websearch_cmd
 from . import url_cmd
+#from . import bibtex_cmd
```
omangin: Should we keep that?

pubs/repo.py
```diff
@@ -192,8 +192,17 @@ def push_doc(self, citekey, docfile, copy=None):
...
+    def unique_citekey(self, base_key, bibentry):
+        """Create a unique citekey for a given base key.
+
+        :param base_key:  the base key in question.
+        :param bibentry:  the bib entry to possibly generate the citekey.
+        """
+        # can't have `/` in citekeys
+        # FIXME: a bit crude, but efficient for now (and allows unicode citekeys)
+        # TODO:  check that the generated citekey does not have a slash too.
+        if '/' in base_key:
```
omangin: Maybe we could export that into a `is_citekey_valid` helper function since it will likely be used elsewhere. There is already a TODO in rename for validating citekeys and I guess this won't be the last one!

benureau:
Ok, I updated the `.travis.xml` file to test the installation. Note that none of those install tests pass right now, as either readme is not included, or the `pubs --help` command fails because there is no config. Those are fixed in the committed code, and the check should pass when we release it to pip.

benureau:
I also reduced the number of online tests to avoid 403 errors from ottobib.com

omangin:
Excellent!

On thing though: I am not sure it makes sense to run the install job every time we push a commit. Building master every once in a while does not hurt on the other side. So I added some conditions to only do the install builds on cron jobs. That way the test should pass here.

benureau:
Nice with the cron jobs. We should release this as 0.8.2, even is the source changed a bit earlier.

PR #184, opened 2019-01-10
try to download pdf of arxiv papers

benureau:
**Very WIP**: just the first working prototype. Still, right now, works, and useful.

TODO:
- [ ] robust requests (no connection, 404, 403, etc.)
- [ ] add option to deactivate behavior
- [ ] clean-up complexification induced in add_cmd code.
- [ ] tests.

One question I have is: what should be the name of the pdf url in the bibtex. Can you think about anything better or more standard than 'pdfurl'?


nbiederbeck:
don't really know the status of this implementation, but I just found out about this: https://github.com/lukasschwab/arxiv.py

PR #185, opened 2019-01-15
Fix/177 convert latex to unicode before printing one-liner

omangin:
Fixes 177: convert latex to unicode before printing one-liner.

The PR actually introduces a new method on the paper object to return a copy of the bibdata which entries' latex have been converted to unicode.

The test currently do not pass due to https://github.com/sciunto-org/python-bibtexparser/issues/225. Some of them will still have to be fixed also with the fixed bibtexparser.

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -19,6 +19,7 @@ six
 
 # those are the additional packages required to run the tests
 pyfakefs
+certifi
 ddt
 mock
 pytest # optional (python setup.py test works without it), but possible nonetheless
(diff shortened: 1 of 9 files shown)


benureau:
Ok, I pushed a small fix for #187 on this PR. It now seems that there are some problems with some test and unicode on 2.7 (one more reason to drop support). Otherwise the PR seems solid.

omangin:
I have just fixed the tests. It was a stupid mistake from my side.

omangin:
BTW, is your fix backward compatible? If not we need to bump the dependency.

benureau:
@omangin: checked with bibtexparser 1.0.1 and pyparsing 2.3.0 on my machine. All good.

Issue #68, opened 2017-07-05
add version control to pubs library


beuerle:
So far it is possible to keep one's pubs library in version control by initiating a git repository and keeping track of changes of the individual files in `bib`, `meta` and `notes` manually by hand (ignore files in `doc` if preferred).

Feature request: add version control possibility

* configure VCS in `.pubsrc` (git, hg, bazaar, ...)
* configure behaviour in `.pubsrc`
	* include `doc`
	* ...
* move `.pubsrc` to config folder (`.pubs/`) to keep track of config changes?
* git: use [`GitPython`](http://gitpython.readthedocs.io/) (or another library?) to implement version control
* make an auto-commit whenever a new reference is added, or an existing reference is modified or deleted

omangin:
Thanks @beuerle. That's indeed a behavior that we have thought about including as a plugin but chose not to implement until now. It is actually a good example of what plugins are for since this does not belong to the core functionalities of pubs.

The road to implement the plugin however requires logging the various events (basically all modifications to data or documents) at the level of the repository, potentially through a _hook_ system defined on these events. This would also benefit a _log_ plugin to write these events to a file but requires a bit of work. This is not our priority right now but if you want to contribute the functionality, we'll be happy to merge it. If others are interested in this, please also let us know.

In the meanwhile, you can either commit manually or use a program such as https://github.com/nesquena/gitdocs to monitor the files and commit on changes.

beuerle:
FWIW, I just configured an alias in `pubsrc` for that for now:

```
cm = !git add . && git commit -a -m autocommit && git push
```

zamlz:
Just want to say that I'm really diggin this project. Of all the cli bib apps i've tried, I think this one seems to fit with what I want it to do. I've only added a couple papers as of now, but if I really like this, I'll move my thing from zotero to pubs. I also would really like git functionality and wouldn't mind looking in how to do this in the future. Hopefully this winter break.

zamlz:
Have you guys seen the program [password-store](https://www.passwordstore.org/)? I was thinking of maybe implementing it similar to this. All commands should automatically make commits to the repo, but if the user wants to do other stuff to their pubs folder, then can just do,
```
pubs git $GIT_CMD
```

omangin:
Yes, something like that would be great!

[...]

PR #191, opened 2019-01-25
Adding git version control as pubs plugin

zamlz:
So here I what I have so far. I haven't worked on it since I last mentioned it 2 weeks ago (as of posting this) in #68 . Like I mentioned, there is still much to fix, but we can move the discussion here.

_I'll restate what I said in the issue #68  for clarity._

Using subprocess respects all of the user settings in the `.gitconfig` file. So far adding, removing, renaming documents are committed to the git repo correctly as far as I'm aware. My next step is probably adding more listeners for other things like modifying notes, tags, etc.

The git shell though is the tricky part which needs to be looked at. The ideal behaviour I was thinking of is that the user could directly operate on the git repo with,

```
$ pubs git log --oneline
$ pubs git pull
$ pubs git push
```

So far, `pubs git push` and simple shells commands do work, but when adding more arguments to them like `pubs git log --oneline` it fails. How the shell commands are to be handled I guess is up for debate.


pubs/plugs/git/git.py
```diff
@@ -0,0 +1,94 @@
...
+from pipes import quote as shell_quote
+
+from ...plugins import PapersPlugin
+from ...events import *
+
+
+class GitPlugin(PapersPlugin):
+
+    name = 'git'
+    pubsdir = None
```
omangin: Same as below, instance variable, not class.

pubs/plugs/git/git.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def __init__(self, conf):
+        self.description = "Run git commands in the pubs directory"
+        # Needed for the event listening
+        GitPlugin.pubsdir = conf['main']['pubsdir']
```
omangin: `self.pubsdir`: we want it to be instance specific (although in practice we use the class as a singleton, there is no need to assume that in the class design).

pubs/plugs/git/git.py
```diff
@@ -0,0 +1,94 @@
...
+from ...events import *
+
+
+class GitPlugin(PapersPlugin):
+
+    name = 'git'
+    pubsdir = None
+
+    def __init__(self, conf):
+        self.description = "Run git commands in the pubs directory"
```
omangin: This one could be a class variable.

pubs/plugs/git/git.py
```diff
@@ -0,0 +1,94 @@
...
+        # Needed for the event listening
+        GitPlugin.pubsdir = conf['main']['pubsdir']
+
+    def update_parser(self, subparsers, conf):
+        git_parser = self.parser(subparsers)
+        git_parser.set_defaults(func=self.command)
+
+    def parser(self, parser):
```
omangin: Could be merged with the previous one since there is no real difference between both.

pubs/plugs/git/git.py
```diff
@@ -0,0 +1,94 @@
...
+        self.parser = parser
```
omangin: I do not see any need to store this here. Do I miss something?

pubs/plugs/git/git.py
```diff
@@ -0,0 +1,94 @@
...
+        p = parser.add_parser(self.name, help=self.description)
+        p.add_argument('arguments', nargs='*', help="look at man git")
+        return p
+
+    def command(self, conf, args):
+        """Runs the git program in a shell"""
+        for a in args.arguments:
```
omangin: Probably forgotten debug print...

pubs/plugs/git/git.py
```diff
@@ -0,0 +1,94 @@
...
+            print(a)
+        GitPlugin.shell(' '.join([shell_quote(a) for a in args.arguments]))
+
+    @classmethod
```
omangin: When `pubsdir` is an instance variable this needs to be a regular instance method.

pubs/plugs/git/git.py
```diff
@@ -0,0 +1,94 @@
...
+        subprocess.call('git -C {} {}'.format(cls.pubsdir, cmd), shell=True)
+
+
+@RenameEvent.listen()
+def git_rename(RenameEventInstance):
+    new_key = RenameEventInstance.paper.citekey
+    old_key = RenameEventInstance.old_citekey
+
+    # Stage the changes and commit
+    GitPlugin.shell("add \*/{}.\*".format(old_key))
```
omangin: Once you have switched to instance variables use `GitPlugin.get_instance().shell(...)` here.
Same for all the following.

omangin:
Thanks @zamlz for posting your work here. I have made a couple of minor comments on your code (see them above).

Regarding the general design I would prefer having fewer events that carry more information instead. Typically instead of the tag, doc, edit and note events I would use a single `PaperEvent` from which `AddEvent` and `RemoveEvent` would be subclasses. We could then add a `description` field to the event that is used for the commit message since the rest of the handling is the same. That would avoid having several different calls for handling each event type (only `PaperEvent` and `RenameEvent` would have to b dealt with, the latter being a sub-class of the first for each only an additional `git add` needs to be performed on the old key (we can offer a unified interface with `PaperEvent` by defining an `@property` that returns `paper.citekey` for the attribute `citekey`)).

This last part was not clear in the original design of events so feel free to implement it or not. I can do it after this PR is merged if not.

In addition to the comments above we would need some basic checks for the new behavior before merging (like a use case where the `git log` is checked after a few papers are added). But this should not be too much work.

pubs/plugs/git/git.py
zamlz: OH, i just realized looked into how get_instance works. That makes a lot more sense now. Thats a really smart way of getting saved instances.

pubs/plugs/git/git.py
zamlz: :+1:

zamlz:
Thanks for reviewing my code @omangin. I've pushed the commit that introduces proper use of the instance variable instead of using the class variable. To be honest, I didn't know what I was doing when I wrote the parser code. I really just copied it form the alias plugin because there was a lot going on there. I'll clean in up later and push the commit for that soon.

I also agree that we should use a single `PaperEvent` and `RenameEvent`. I'll try to see if I can set those up as well.

omangin:
Great! Let me know when you feel it is ready for a second review.

zamlz:
@omangin Sorry for the late followyup, what you elaborate more on the way the `PaperEvent` parent class works

omangin:
Sorry, my message was not so clear and what I meant included some re-design of the event classes. Right now the `PaperEvent` class does not exist. I was thinking of having something like the following.
```python
class PaperEvent(Event):
    _format = "Unknown modification of paper {citekey}."

    def __init__(self, citekey):
        self.citekey = citekey

   @property
   def description(self):
        return self._format.format(citekey=self.citekey)


class RemoveEvent(PaperEvent):
    _format = "Removes paper {citekey}."


class ModifyEvent(PaperEvent):
    _format = "Modifies paper {citekey}."


class RenameEvent(PaperEvent):
    _format = "Renames paper {old_citekey} to {citekey}."

    def __init__(self, paper, old_citekey):
        super(RenameEvent, self).__init__(paper.citekey)
        self.paper = paper
        self.old_citekey = old_citekey

  @property
   def description(self):
        return self._format.format(citekey=self.citekey, old_citekey=self.old_citekey)
```
That way you only need a single `@PaperEvent.listen()` inside which you add an additional add (!) when it is passed a `RenameEvent` (you need to add the removal of the old files). You can use the `event.description` as a commit message.

pubs/plugs/git/git.py
```diff
@@ -0,0 +1,44 @@
+
+import subprocess
+from pipes import quote as shell_quote
+
+from ...plugins import PapersPlugin
+from ...events import *
```
omangin: Avoid wildcard imports. Always make them explicit (otherwise it is hard to tell what comes from where and you may have namespace collisions).

omangin:
See #194 for broken test (you can merge or rebase on master once it is merged).

zamlz:
Thanks @omangin. I'll let you know when I finish getting through your suggestions. On a side note, how should we handle the `docsdir`? Judging form the `pubsrc` file, it seems like it could exist anywhere. Should I be adding those files to the git repository by default? Or should it be a setting in the config file that the user can toggle. In my personal papers repository, I've been storing the pdfs as well because I don't expect them to change, but it may be different. In any case, the yaml file will get updated so that needs to be version controlled each time a document is added. Would like to know your thoughts on this.

omangin:
Yes, this could be an issue but as far as I am concerned and, I guess for most users, the desired behavior is to commit the documents, which is also the current one.

For me the `.gitignore` seems the perfect way to solve this and it does not require any new code on our side! The more we can rely on git functionalities to keep the pubs side of code lightweight, the better it is.

If the option is really needed later then we can add it. Because currently the plugin just does a wildcard add and is unaware of the various sub-directories, this would go well with another update which would add all files individually but would also take a little more design work. Since the first version likely already fits many use cases, I believe a good strategy is to ship it and then we will better see what is missing from usage.

benureau:
Nice work @zamlz. If the doc directory is not inside the pubs directory (in another dir or even volume), the files cannot be added to the git. It might be worthwhile to notify the user of this in the documentation of the functionality at some point. And you should only care about docsdir when it is `docsdir://` (btw, this is confusing in the configuration. `pubsdir://doc/` would be more straightforward).

benureau:
@zamlz: oh, and don't worry one second about delaying your contribution when you have more important stuff to do. It's the same for us, actually. In fact, it's right there in the readme: _"We may not answer right away (science comes first!) but we'll eventually look into it."_

omangin:
> (btw, @omangin, in the configuration. `docsdir = docsdir://` is confusing, don't you think? `docsdir = pubsdir://doc/` would be more straightforward, no?).
Yep! Good point.

zamlz:
So far, all I've done is add more events. I'll get to creating the unit tests when I can though.

pubs/plugs/git/git.py
```diff
@@ -0,0 +1,44 @@
...
+    if isinstance(PaperEventInstance, RenameEvent):
+        old_citekey = RenameEventInstance.old_citekey
+    else:
+        old_citekey = None
+
+    # Stage the changes and commit
+    git = GitPlugin.get_instance()
+    if old_citekey:
+        git.shell("add \*/{}.\*".format(old_citekey))
+    git.shell("add \*/{}.\*".format(citekey))
```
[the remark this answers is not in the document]
omangin: This is not outdated.

omangin:
> So far, all I've done is add more events. I'll get to creating the unit tests when I can though.

It is really close to merge-state now! 

Actually the testing approach I described in https://github.com/pubs/pubs/pull/191/#issuecomment-457972880 won't work here because we cannot run `git` in the fake filesystem that is used for the test. The way to go would have to mock the `subprocess` call to check its content instead. 
However, since this is slightly more complex than the other solution, so feel free to let us know if you would  rather have us handle it.

Please also have a look at the comments [261483624](https://github.com/pubs/pubs/pull/191/#discussion_r261483624) and [r261484969](https://github.com/pubs/pubs/pull/191/#discussion_r261484969): they should be pretty straightforward to address!

zamlz:
@omangin could you explain what you mean in [261483624](https://github.com/pubs/pubs/pull/191/#discussion_r261483624)? I've added escaping for the special characters you mentioned in the other comment. 

As for the testing method, I think it would be best if I left that part to you guys. I don't think I would have the time to properly figure it out in a reasonable amount of time.

zamlz:
@omangin Hey I know you must be busy, but just wanted to make sure this project is still alive ;)

benureau:
@zamlz: It's true, that recently, we may wonder. Yes. I actually will get some time next week to take care of pubs. I plan to tackle adding the tests and merging your PR first.

zamlz:
Let me know if there anything else you need me to do on my end ;)

benureau:
@zamlz: done! I ended up having to copy/write a significant amount of code for the tests. Once I add those running, I had to make some changes and additions to the plugin. I opened a new pull request with my commits on top of yours (#197), and I sent you an invitation to be a collaborator on the project with read access, so you can review my changes. I'd like to have your input and possibly could you test the code in your repo a bit before we merge? I am doing the same on my end.

I am closing this, we can follow this up in #197.

zamlz:
Thanks! I'll get around to testing this soon and give my feedback in the new pull request.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #193, opened 2019-02-20
Add CitekeyFilter and FIELD_ALIAS 'key': 'citekey' to query.py

shanewstone:
Thank you for writing and maintaining `pubs`!

This PR is a simple addition to `query.py` which enables users to query the publication database using citekeys. This is useful because the citekey is the unique identifier used to interact with each publication, e.g. through `pubs doc open`. Citekey queries are different than author queries, since a citekey query can achieve more specificity in just a few characters: as an example, `pubs list citekey:Fox201` returns all first-authored papers by Fox in the 2010s. Further, I keep track of papers in my head by first author and year, so this just seems natural to me and I would assume it would feel natural to others as well.

As part of the change, I added `key` as an alias for `citekey` to enable shorter commands like `pubs list key:Fox201`.

(Note that the general `FieldFilter` does not currently work for citekeys since `FieldFilter` only looks through `paper.bibdata`.)

### DIFF
--- a/pubs/query.py
+++ b/pubs/query.py
@@ -15,6 +15,7 @@ FIELD_ALIASES = {
     't': 'title',
     'tags': 'tag',
     'y': 'year',
+    'key': 'citekey',
 }
 
 
@@ -78,6 +79,12 @@ class AuthorFilter(QueryFilter):
                         for author in paper.bibdata['author']])
 
 
+class CitekeyFilter(QueryFilter):
+
+    def __call__(self, paper):
+        return self._is_query_in(paper.citekey)
+
+
 class TagFilter(QueryFilter):
 
     def __call__(self, paper):
@@ -137,6 +144,9 @@ def _query_block_to_filter(query_block, case_sensitive=None, strict=False):
     field, value = _get_field_value(query_block)
     if field == 'tag':
         return TagFilter(value, case_sensitive=case_sensitive, strict=strict)
+    elif field == 'citekey':
+        return CitekeyFilter(value, case_sensitive=case_sensitive,
+                            strict=strict)
     elif field == 'author':
         return AuthorFilter(value, case_sensitive=case_sensitive,
                             strict=strict)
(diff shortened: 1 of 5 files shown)


shanewstone:
I have never written unit or integration tests before, but the idea was clear enough in [test_queries.py](https://github.com/pubs/pubs/blob/d626fe4251836c59b0c84091a95b8844c9289e93/tests/test_queries.py) that I was able to write 7 unit tests that make sense to me: one to check for query failure if there is no citekey, two to test proper case matching, one for checking proper LaTeX encoding, one for proper unicode normalization, and two for testing the use of `--strict`. I borrowed heavily from the unit tests already present. I ran `python setup.py test` and everything checks out.

For the integration tests, and since I have not done this before, you may have to hold my hand a little bit: I do not know what kind(s) of tests of a new query filter would be appropriate for [test_usecase.py](https://github.com/pubs/pubs/blob/master/tests/test_usecase.py#L421).


Thanks!

pubs/query.py
```diff
@@ -137,6 +144,9 @@ def _query_block_to_filter(query_block, case_sensitive=None, strict=False):
     field, value = _get_field_value(query_block)
     if field == 'tag':
         return TagFilter(value, case_sensitive=case_sensitive, strict=strict)
+    elif field == 'citekey':
+        return CitekeyFilter(value, case_sensitive=case_sensitive,
+                            strict=strict)
```
omangin: Minor: indentation.

pubs/query.py
```diff
@@ -15,6 +15,7 @@
     't': 'title',
     'tags': 'tag',
     'y': 'year',
+    'key': 'citekey',
```
omangin: Maybe we could also add an example to the `QUERY_HELP` above.

shanewstone:
Thanks for the help! I think everything is ready now.

I noticed that `import mock` appears twice at the top of `test_usecase.py`.  I didn't change it, but thought I would point it out.

PR #198, opened 2019-05-09
ENH: Don't fail on import, warn instead

ksunden:
Citekeys that were already in the repo failed and halted execution.
This means that what to me seems like a common case of importing a `.bib`
file that contains both previously entered and new entries would fail or require
`-O` (which would overwrite)

I had tried `--ignore-malformed` to make sure that didn't handle this case, but it still failed.

### DIFF
--- a/pubs/commands/import_cmd.py
+++ b/pubs/commands/import_cmd.py
@@ -104,7 +104,14 @@ def command(conf, args):
     keys = args.keys or papers.keys()
     for k in keys:
         p = papers[k]
-        rp.push_paper(p, overwrite=args.overwrite)
+        try:
+            rp.push_paper(p, overwrite=args.overwrite)
+        except repo.CiteKeyCollision:
+            ui.warning("{} already in repository, use '-O' to overwrite".format(
+                    color.dye_out(p.citekey, 'citekey')
+                )
+            )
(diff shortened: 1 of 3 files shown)


benureau:
@ksunden: Thanks for the contribution! It seems indeed a nice improvement on the behavior of the import command. I'll formally review your contribution soon.

ksunden:
Looks like there is a test that explicitly tests for exit in this case. I will track that down later tonight or so.

(I was just importing a .bib file last night and wanted this behavior, so I made sure it worked in my usage)

ksunden:
Okay, pushed an updated test. Rather than testing that the command fails, I actually look at the output of `pubs list` to ensure that the entry is the original.

To do this, I added a new string fixture that has the same citekey as `Page99`, but a different year, thus the output of `list` ends with "(1999)", but would be different if it was overwritten.

It's not the cleanest of tests I can imagine, but it seems in line with the the rest and is not so bad.

Let me know if you have other ideas for testing/want to think up some other way of testing for this behavior.

ksunden:
Noticed after I did this that there are some tests which do similiar changes using re substitutions, rather than creating a new base string. If you would prefer that method, I can do it that way.

PR #199, opened 2019-06-25
adds chronological listing using bibtex year field

emnllm:
This small change adds the `-C` / `--chronological` option to the `list` command in order to output the entries based on the year of publication.

### DIFF
--- a/pubs/commands/list_cmd.py
+++ b/pubs/commands/list_cmd.py
@@ -48,8 +48,8 @@ def command(conf, args):
         papers = [p for p in papers if p.docpath is None]
     if args.alphabetical:
         papers = sorted(papers, key=lambda p: p.citekey)
-    if args.chronological:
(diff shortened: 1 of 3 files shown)


benureau:
Thanks @emnllm! That seem quite a relevant feature. Could you add one relevant test to automatically check that 1) future code change don't break this and 2) it actually does what it is suppose to be doing on a simple case? You can copy and modify an existing test from there: https://github.com/pubs/pubs/blob/eaeb48bd206d3525bcee5dcc4d2f49bf9c3ba58c/tests/test_usecase.py#L420

emnllm:
The new commits:
- handle the absence of a year field by printing these papers at the end, sorted by `date_added`
- add a test to cover the normal case and the exception above

Possible improvement:
- handle a sub-sorting by month: it would require some parsing of the different types of month representations that are used (month name, abbreviation, numerical representation, translations?)

Issue #161, opened 2018-08-14
Formatting with `pubs list`


benureau:
The `&` and accents are not well displayed:

```
$ pubs list
[Watson2016] Watson, Richard A. and Szathm\'ary, Eörs "How Can Evolution Learn?" Trends in Ecology \& Evolution (2016)
```

natj:
I would also consider aligning different fields, i.e.:

```
[citekey1]         Sironi, L. and Spitkovsky, A. "Efficiency of Particle Acceleration in Relativistic Magnetized Shocks" AAS/High Energy Astrophysics Division \#11 (2010)
[longer_cite_key2] Klimontovich, Yu L            "The Statistical Theory of Non-Equilibrium processes in a plasma" Journal of Plasma Physics (1969)
[another3]         Esirkepov, T. Z.              "Exact charge conservation scheme for Particle-in-Cell simulation with an arbitrary form-factor" Computer Physics Communications (2001)
[ck4]              Jackson, J. D.                "Classical electrodynamics" (1975)
```

P.S. Nice work btw, this was exactly what I was planning to write myself too!

natj:
Furthermore, I was toying with the idea of adding different "global" sort options. For example 

- `--sort=tags` group different entries with a same tag together (can print same citekey multiple times)
- `--sort=author` sort alphabetically by author name
- `--sort=year` sort chronologically by publication year

These need a little bit of work to work nicely with the current options but for example the tag option sounds quite useful to me.

omangin:
Thanks for the feedback!

Hopefully #185 should fix the initial issue.

I like the `--sort` option idea. It is actually already discussed as #109 and just waiting for an implementation! So a contribution would be welcome! (I do not have an opinion toward naming the option `sort` or `order`. Both are fine.)

Personally I am not a big fan of the tabular display by default but I understand that it can be useful and that this is also a matter of taste. Customizing the output also is a request that comes regularly so I would rather implement this by enabling user customization than just implementing this one feature. But that is open to discussion and a nice PR can be very convincing in favor of merging the functionality ;)

zamlz:
I think it could be cool if the user could specify a format string in the `.pubsrc` file. Maybe something like this:
```
[banino2018] Banino, Andrea et al. "Vector-based navigation using grid-like representations in artificial agents" Nature (2018) [pdf] | usc/resl/rg18
```
would be
```

list_format = '[%citekey%] %authors% "%title%" %journal% (%year%) %metadata%'
```
[...]

PR #205, opened 2019-08-07
Support for List Formatting

zamlz:
As I mentioned in #161, I was gonna give this a go. Sorry it took so long, I only got started on it like two weeks ago.

Back to topic, this is an incomplete PR. It still needs some work and some tests which I'll write once I get some questions answered. Lets first talk about what this PR sets out to do. Buckle in, because this might be a long read.

## Goal

Basically, we want the ability for the user to provide a format string so that they can customize the output of `pubs list`. The idea is that this would be given in the config file. The default one would look something like this.
```
list_format = '[$citekey] $authors "$title" $journal ($year) $pdf | $tags'
```
In turn this would produce the output, 
```
[banino2018] Banino, Andrea et al. "Vector-based navigation using grid-like representations in artificial agents" Nature (2018) [pdf] | usc/resl/rg18
```

The reason for the `$variable` syntax is because its the we use python string Templates as @omangin suggested. The primary reason for this is for security. If we used python's `.format()` function, we could access variables and things we aren't suppose to.

At this point, the basic list formatting for `pubs list` was complete. But I felt that it needed a couple more features. Specifically, the ability to apply conditionals parts to the string format and to apply padding features. To implement this, I introduced a new syntax to the Template class. Basically, here is how it works.

### Format Blocks

`{value:format?condition}` is defined as a block. Basically `format` works the same way as python format should work (for applying padding and things like that). `value` is whatever string value you wish to format. `condition` is the conditional expression when evaluated to non-empty (I'll explain that in a second), we use value as is, if it is empty, then we replace the whole block with the empty string. Here is an example illustrating this.

### Condition Expressions

Suppose this is the value of my `list_fmt` in my config.
```
list_fmt = "[$citekey]: $title {($year):?$year}\n    authors: $authors{\n    journal: $journal:?$journal}{\n    meta:   :?$ext$tags}{ [$ext]:?$ext}{ |:?$ext,$tags}{ $tags:?$tags}\n"
```
This would be the following output I would get
```
[sutskever2014]: Sequence to Sequence Learning with Neural Networks 
    authors: Sutskever, Ilya et al.
    meta:    [pdf] | seq2seq,usc/csci699/seq2seq

[frankle2019]: The Lottery Ticket Hypothesis: Finding Sparse, Trainable Neural Networks (2019)
    authors: Frankle, Jonathan and Carbin, Michael
    journal: International Conference on Learning Representations
    meta:    [pdf]
```
Notice how the the journal line is not there anymore. This is because of 
`{\n    journal:$journal:?$journal}`. Basically `\n    journal:$journal` is only printed if and only if the condition `$journal` is non-empty. Since there isn't a value for journal in the first paper that is in the output, it doesn't have have that string printed out. But the second paper does since there is a journal value.

Lets look at a more complex example. Also in the format string is the substring `{\n    meta:   :?$ext$tags}{ [$ext]:?$ext}{ |:?$ext,$tags}{ $tags:?$tags}`. Here is four blocks in quick succession with the goal of printing out metadata. The first block decides whether to print out metadata information or not. The condition is `$ext$tags`. `$ext` is basically the doc extension and `$tags` is paper tags. The condition `$ext$tags` is basically an OR statement. Meaning if either the extension value or tags is non-empty, we display this value. The next block performs the same time of conditional as the journal example but for the document extension so I'll skip it. The next conditional is `$ext,$tags` which is an AND statement which checks to see if both extension and tags are are non-empty strings. Only then will it display the seperator. The last block is again similar to the journal example. One last implementation detail is order of precedence. Something like `?$foo$bar,$var` would equate to `($foo OR $bar) AND $var`. If you have any other questions or issues with the formatting, please let me know. I was also tihinking about how to incorparate a NOT operation in there, but I didn't really see a real reason to implement it. Maybe we should add it in the interest of being logically complete.

### Formating Options

This is much simpler to explain than the latter. The format still follows the same syntax as python formats. However, we replace in such a way that should be safe. The code should explain that part. It follows this procedure.
```python
if fmt != ':' :
    val = ('{' + fmt + '}').format(val)
```
Instead of evaluating `val` which is unsafe, we simply substitute it in.

### Problems

As you may have realized, there are some problems with this approach.
- Commas present in the values of the variable strings could potentially break the conditional syntax. (possible solution in https://github.com/pubs/pubs/pull/205#issuecomment-519777414)
- The regex used for finding the blocks is pretty good, but I would like to prove somehow that is going to work 100% of the time  as intended with either tests or something.
- The workaround for getting padding (and other formatting as provided by python format) seems like it would be safe, but I'm unsure if it truly is.
- Padding issue with color entities (https://github.com/pubs/pubs/pull/205#issuecomment-519777414)

Sorry for the really long PR. I just wanted to make sure you guys were caught up in my thought process. Any comments/improvements would be appreciated. If you guys think this syntax is good, I can go ahead and start making tests to make sure that it does work as intended. Though I imagine the conditional syntax may need to be changed in some way.


benureau:
Hi @zamlz. That seems terrific work. I'll get a chance to get to look at it in the next few days.

zamlz:
Thanks @benureau, looking forward to it!

Couple things that might be better to change after you've done your initial review. Instead of using commas in the conditional expression might not be good since there could be commas in the individutal entities themselves which could destroy the original meaning of the expression. It might be better to use `|` instead to seperate them since I don't see how `|` would appear in string values of a bibtex file (unless there is some really weird edge case I'm not thinking of).

Another problem I've found is that suppose an entity like `$citekey` wants some padding. We could say it has like 10 space padding on the left.  If the `$citekey` was 6 characters long, the intended result would be to expect that there are 4 spaces in front of it. However, thats not what happens. `$citekey` doesn't just contain the citkey, but all the hidden characters that represent color. As a result, you need to specify that you want to pad like 15 spaces (random guess) if you want it to visually pad 4 spaces . The solution for this is to probably figure out some way to trim the color information and reclip it back on after the string has been padded. I'll look into that too.

benureau:
Ok, I have looked through your proposed changes (haven't run them though), here are my remarks:
* condition syntax: as you mention, it may be problematic. Also, why so concise? It is far from intuitive. Why not, rather than `$foo$bar,$var` actually have `($foo or $bar) and $var`?
* you should expect that any character can show up in bibstrings. Since you only test for absence/emptyness, you don't need to substitute the conditions. You probably want to evaluate them explicitly, with a boolean for each value of the metadata. That will probably require to substitute block by block. 
* for the color issue, indeed. We probably need a modified version of undye that returns the color as well so we can reapply it.
* you may want to catch exceptions, especially when substituting. It should probably fail gracefully, with reverting to the one-liner function if anything goes wrong and displaying a warning.
* I guess some tests indeed could make us understand how it all works and what are the kings. 

Overall, I feel the system you designed is quite powerful, and is on the verge of being too complex. In the end, we need a solution that is robust and easy to use. If a functionality is too complex, only power users will use it. In that case, we can make it a plugin, to clearly signify that regular users don't have to deal with it until they feel the need.

zamlz:
Thanks @benureau,

Yes, ideally that would be a nice conditional syntax, but the reason I chose this concise form was because it was easy to write the code for parsing it and less characters to write out in the config file. Let me try experimenting with that. I also agree that I should evalutate them explicitly. I think I have a solution for the color issue, and hopefully I'll have that for the next review. Yes, I don't want this syntax to be too complex for the end-user. Hopefully that will not be the case. Thanks again for the review.

zamlz:
Hey @benureau, sorry I haven't updated this in a while. Been very busy as of late in my personal life. I have thought a about this problem when I do get the chance though. The issue is supporting the formatting options. I can get rid of that and would be able to finish this PR. However, if we really do wish to keep the formatting functionality, I might not be able to do it with the given time I have (or its going to take quite a while). It's going to require a massive code rework I think because we need to rework the color system in some way that respects the formatting. I don't really know of any easy way to do this. Let me know which option you prefer.

benureau:
Hi @zamlz. No problem about the delay. I am often in the same boat. It's not urgent, and if you want to do it right, then let's do it right, even if it takes one, three, six months. No pressure.
Careful about massive reworks though; when you've find a solution that might work but that will shift many things around, drop me one or two lines about it so we can discuss it before you start implementing.

zamlz:
Doing some housecleaning so I'm closing this repo. Unsure if I'll ever get back to this.

benureau:
Ok, thanks for notifying us, and thank you for the work you have done, @zamlz.

zamlz:
Thank you to you too for providing this app! Helped me a bunch that's for
sure!

PR #209, opened 2019-12-09
Use generated publication key as default when adding a new publication

jkulhanek:


zamlz:
Perhaps we could have both functionalities? I imagine some people want the
citekeys that are obtained from the bib file.

Like we could use a simple command line flag that toggles tells pubs to use
the generated one instead? Not sure what we call this flag though. We could
also have a setting in the pubrc file that hard toggles this.

jkulhanek:
Sure, It would be great to allow for setting the default behavior in the config and also using a flag. What about ‘normalize_citekey’?

zamlz:
Thats a good name for the operation! Look forward to seeing the changes.

Incidentally, I've been manually changing the citekeys this whole time to
the lowercase of the last name of the first author + the year of
publication. This would be a good feature to have for sure.

jkulhanek:
What do you think about these changes? Should we add some tests for the code?

pubs/bibstruct.py
```diff
@@ -72,8 +72,16 @@ def generate_citekey(bibdata):
...
+        raise ValueError(
+            "No author or editor defined: cannot generate a citekey.")
+    try:
+        first_word = entry['title'].split(' ')[0]
+    except KeyError:
+        first_word = ''
+    author_last_name = author_last(first_author)
+    citekey = format_string.replace('%Y', year).replace('%y', year[-2:]) \
+            .replace('%A', author_last_name).replace('%a', author_last_name.lower()) \
+            .replace('%W', first_word).replace('%w', first_word.lower())
```
zamlz: Instead of calling replace a bunch of times perhaps we could use [Python Template Strings](https://docs.python.org/3.6/library/string.html#template-strings)?

jkulhanek:
Changed to support python's format syntax. I also added tests for pubkey generation.

jkulhanek:
Any update on this?

pubs/bibstruct.py
```diff
@@ -1,10 +1,19 @@
 from __future__ import unicode_literals
+try :
+    import __builtin__
+except: 
+    # Python 3.x
+    import builtins
+    if 'unicode' not in builtins.__dict__.keys():
+        unicode = str
```
benureau, line 8: Actually not needed, the p3 module provide ustr, which does the job

pubs/bibstruct.py
```diff
@@ -56,26 +65,51 @@ def valid_citekey(citekey):
     # FIXME: a bit crude, but efficient for now (and allows unicode citekeys)
     return not '/' in citekey
 
-
-def generate_citekey(bibdata):
+class CitekeyFormatter(Formatter):
```
benureau, line 68: Nice use of `string.Formatter`. Lacks a bit of documentation.

pubs/bibstruct.py
```diff
@@ -56,26 +65,51 @@ def valid_citekey(citekey):
...
+    def get_value(self, key, args, entry):
+        if isinstance(key, (str, unicode)):
+            okey = key
+            if key == 'author' and not 'author' in entry:
+                key = 'editor'
+            elif key == 'editor' and not 'editor' in entry:
+                key = 'author'
+
+            if key == 'first_word' and 'title' in entry:
+                return entry['title'].split(' ')[0]
```
benureau, line 93: A bit brittle. Won't work properly on title with colons, commas (e.g. "Long, Windy Title", "Robots: the Next Revolution").

pubs/bibstruct.py
```diff
@@ -56,26 +65,51 @@ def valid_citekey(citekey):
...
+            fmt = fmt[1:]
+        elif len(fmt) > 0 and fmt[0] == 'l':
+            s = str(val).lower()
+            fmt = fmt[1:]
+        else:
+            s = val
+        return str2citekey(s.__format__(fmt))
+
+
+    def get_value(self, key, args, entry):
```
benureau, line 84: One likely problem: content does not get sanitized. So names with latex in them may not produce legal citekeys (i.e. "Le~Clerc de La~Th{\'e}venaz"). Sanitizing here might not be the best way though, as it might be hard. Better would be to check and gracefully fail, and inform the user that she has to edit the citekey herself. In any case, usecases are the way to go to test robustness against those.

pubs/bibstruct.py
```diff
@@ -56,26 +65,51 @@ def valid_citekey(citekey):
...
+        if isinstance(key, (str, unicode)):
+            okey = key
```
benureau, line 86: Being less terse is useful here: `original_key` instead of `okey` would help the next programmer understand faster.

pubs/config/spec.py
```diff
@@ -31,6 +31,25 @@
...
+# the form '{{substitution:modifier}}' and '{{substitution}}' will 
+# be substituted with their appropriate values. The following
+# substitutions are used:
+#    author_last_name: last name of the first author
+#    year: year of publication
+#    first_word: first word of the title
+# modifiers:
+#    l: converts the text to lowercase
+#    u: converts the text to uppercase
+# examples:
```
benureau, line 47: Some example using regular entries could be nice, such as `journal` or `month`.

pubs/config/spec.py
```diff
@@ -31,6 +31,25 @@
...
+normalize_citekey = boolean(default=False)
+
+# String specifying how to format the citekey. All strings of 
+# the form '{{substitution:modifier}}' and '{{substitution}}' will 
+# be substituted with their appropriate values. The following
+# substitutions are used:
+#    author_last_name: last name of the first author
+#    year: year of publication
+#    first_word: first word of the title
+# modifiers:
```
benureau, line 44: I think an important modifier that would increase flexibility and automation would be "optional". Include the field if found. Could be used for adding the month, when present, without auto-generation fail on half the entries.

pubs/config/spec.py
```diff
@@ -31,6 +31,25 @@
...
 debug = boolean(default=False)
 
+# If true the citekey is normalized using the 'citekey_format' on adding new publications.
+normalize_citekey = boolean(default=False)
+
+# String specifying how to format the citekey. All strings of 
+# the form '{{substitution:modifier}}' and '{{substitution}}' will 
+# be substituted with their appropriate values. The following
+# substitutions are used:
+#    author_last_name: last name of the first author
```
benureau, line 41: Because it can also be the name of the editor, I think `last_name` is better. Also, the editor possibility needs to be documented.

tests/test_bibstruct.py
```diff
@@ -32,6 +32,46 @@ def test_simple(self):
         key = bibstruct.generate_citekey(bibentry)
         self.assertEqual(key, 'Salinger1961')
 
+    def test_no_modifier(self):
```
benureau, line 35: Those tests are good. We need, as mentioned above, some usecase tests with tricky cases to ensure the commands work and fail gracefully.

jkulhanek:
Thanks for the review! I think it would be best if you did the changes. It is your codebase and you are definitely more experienced programmer than I am.

benureau:
Alright. I will do those over the course of the next few days.

jkulhanek:
Is there any update?

benureau:
Sorry, I will make some time next week-end to go over this.

benureau:
So, the main issue is to sanitise the input data that generate the key so that it does not create problem down the line. I think a best effort is okay, and should be fine for most use cases, even if it occasionally mangle unicode names. It still needs some work and tests. I will continue to work on it throughout the week.

jkulhanek:
I have updated the default citekey to match Google Scholar citekey generation and merged "main" branch from the main repository into the PR branch.

jkulhanek:
Is there any update?

jkulhanek:
Is there any update?

benureau (approved):

benureau:
I tested it, and everything seems to work. Quality work. Sorry for the late merge.

jkulhanek:
Thank you very much!

Issue #212, opened 2020-01-07
Make number of author before 'et al.' a configuration parameter


cristobaltapia:
I am using pubs together with wofi [0] (like rofi but for wayland), so I basically use the `list` command to get all the papers and then search through them for authors' names or whatever. The problem is that not all authors are present in the string if the paper has more than three authors due to the 'et al.' notation. This number is hard-coded [here](https://github.com/pubs/pubs/blob/master/pubs/pretty.py#L29).

Would you consider making this number a configuration parameter? It should be a fairly easy change.

[0] https://hg.sr.ht/~scoopta/wofi

benureau:
Yes, that should be a parameter in the configuration, and that should be easy to do. We could even have `0` as a special value saying that `et al.` should never be used.

cristobaltapia:
Yes, the `0` parameter is a good idea (much better that what I was thinking: just writing 99 :P )

benureau:
Would not have worked with this one: https://arxiv.org/abs/1109.4897v4 :)

cristobaltapia:
haha no, that one would have not worked. I hope the list of names don't count for the maximum number of pages allowed.

cristobaltapia:
Thanks! I was literally thinking about this yesterday, but saw no update. And then I get the Email notification :P
It is working as expected for me.

PR #224, opened 2020-05-03
Fix #223: update python versions in travis.

omangin:
- remove python version unsupported by pyfakefs from testing in Travis,
- add python 3.8.

PR #225, opened 2020-05-04
Feat/212 n authors

benureau:
Added an option in the config to have more than three authors displayed, as per #212.

### DIFF
--- a/pubs/config/spec.py
+++ b/pubs/config/spec.py
@@ -27,7 +27,8 @@ edit_cmd = string(default='')
 # Which default extension to use when creating a note file.
 note_extension = string(default='txt')
 
-# How many authors to display in
+# How many authors to display when displaying a citation. If there are more
+# authors, only the first author is diplayed followed by 'et al.'.
 n_authors = integer(default=3)
 
 # If true debug mode is on which means exceptions are not catched and
--- a/pubs/config/spec.py
+++ b/pubs/config/spec.py
@@ -27,6 +27,9 @@ edit_cmd = string(default='')
 # Which default extension to use when creating a note file.
 note_extension = string(default='txt')
 
+# How many authors to display in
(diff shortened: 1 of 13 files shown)


benureau:
Rebased on top of PR #224, btw.

pubs/pretty.py
```diff
@@ -23,19 +23,30 @@ def person_repr(p):
...
+            authors_str = '{} et al.'.format(authors[0])
         else:
-            return authors[0] + (' et al.' if len(authors) > 1 else '')
+            authors_str = ' and '.join(authors)
+
+        # if n_authors > 0:
+        #     authors = authors[:n_authors]
+        # authors_str = ' and '.join(authors)
+        # if 0 < n_authors < len(bibdata['author']):
+        #     authors_str += ' et al.'
```
omangin, line 42: We can get rid of this.

pubs/config/spec.py
```diff
@@ -27,6 +27,10 @@
 # Which default extension to use when creating a note file.
 note_extension = string(default='txt')
 
+# How many authors to display when displaying a citation. If there are more
+# authors, only the first author is diplayed followed by 'et al.'.
+n_authors = integer(default=3)
+
```
omangin, line 33: Because I did not read the full line I was initially under the impression that `n_authors` meant "we display n then put 'et al.'". Maybe that's just me but we could also write "Maximum number of authors..." or use `max_` in the variable name. I am not sure, though, what would be the clearest and it is really minor, so you can ignore this comment if you feel differently.

pubs/pretty.py
benureau: Indeed

pubs/config/spec.py
benureau: Yes, `max_authors` is more correct than `n_authors`.

PR #226, opened 2020-05-05
Support for non-standard bibtex

benureau:
Provides support for any non-standard bibtex entry. Quite useful with modern bibtex data.
Closes two issues, rebased on top of PR #225


benureau:
Probably not worth reviewing this one separately from #227. Closing this.

PR #231, opened 2020-05-19
`conf` not defined in `resolve_citekey`, use `self.conf`

beuerle:
With ambiguous citekeys, the following error occurred:

```
$ pubs edit Li
error: Be more specific; 'Li' matches multiples citekeys:
error: name 'conf' is not defined
Traceback (most recent call last):
  File "/usr/bin/pubs", line 11, in <module>
    load_entry_point('pubs==0.8.3', 'console_scripts', 'pubs')()
  File "/usr/lib/python3.8/site-packages/pubs/pubs_cmd.py", line 107, in execute
    if not uis.get_ui().handle_exception(e):
  File "/usr/lib/python3.8/site-packages/pubs/pubs_cmd.py", line 104, in execute
    args.func(conf, args)
  File "/usr/lib/python3.8/site-packages/pubs/commands/edit_cmd.py", line 33, in command
    citekey = resolve_citekey(rp, args.citekey, ui=ui, exit_on_fail=True)
  File "/usr/lib/python3.8/site-packages/pubs/utils.py", line 36, in resolve_citekey
    ui.message('    {}'.format(pretty.paper_oneliner(p, max_authors=conf['main']['max_authors'])))
NameError: name 'conf' is not defined
```

`resolve_citekey` should access `self.conf`, as no `conf` parameter is
given.

### DIFF
--- a/pubs/utils.py
+++ b/pubs/utils.py
@@ -33,7 +33,7 @@ def resolve_citekey(repo, citekey, ui=None, exit_on_fail=True):
                      "citekeys:".format(citekey))
             for c in citekeys:
                 p = repo.pull_paper(c)
-                ui.message('    {}'.format(pretty.paper_oneliner(p, max_authors=conf['main']['max_authors'])))
+                ui.message('    {}'.format(pretty.paper_oneliner(p, max_authors=self.conf['main']['max_authors'])))
             if exit_on_fail:
                 ui.exit()
     return citekey


benureau:
@beuerle: Thank you. Indeed, `conf` is not given to the function. After just a cursory examination, I don't understand why the fonction would have any access to `self.conf` though? It's not a part of a class. I think the best way would be to give a conf argument to `resolve_citekey`. What do you think?

Anyway, this shows that `resolve_citekey` is not appropriately tested in our test suite. I'll write a test to do that soon.

beuerle:
Indeed. I have to ramp up my testing skills.

I'll submit a new PR with `conf`as an argument for `resolve_citekey`.

PR #233, opened 2020-05-26
fix missing conf argument for resolve_citekey

beuerle:
Proper PR instead of #231 .

There is some strange behaviour coming from `ui.error`, though (at least with python `3.8.3`):

The error message `error: Be more specific; 'citekey' matches multiples citekeys:`

appears **after** the list of publications which match the citekey and comes last. This should not be the case, as `ui.error` is called before listing the publications (and indeed, any other `print` command in `ui.error` actually gets executed).

The problem lies with https://github.com/pubs/pubs/blob/master/pubs/uis.py#L88. If `**kwargs` is removed from the `print` statement, the sequence is preserved.

### DIFF
--- a/pubs/commands/doc_cmd.py
+++ b/pubs/commands/doc_cmd.py
@@ -71,7 +71,7 @@ def command(conf, args):
     # ui.exit()
 
     if args.action == 'add':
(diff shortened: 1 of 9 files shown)


benureau:
Added the test, everything is in master.

Issue #234, opened 2020-06-02
pubs tag improvements


benureau:
The tag commands throws error when no matching citekeys are found, and then assume it is a tag. This is confusing. Also, when no papers match, it should be explicit.

benureau:
Additionally, the tags are saved as a set in the metadata, which may not be needed, is difficult to read and edit, might lead to slowdown in parsing and decoding. Storing as a (sorted?) list might be simpler and better.

benureau:
Also, empty tags ('') are valid. This is not desired.

PR #238, opened 2020-06-08
add `--no-tags` argument to list command

### LINKED ISSUE #237: add `--no-tags` to `list` command
Analogous to `--no-docs` this argument would only list publications without any tags.

I might look into this and create a PR.

beuerle:
This argument works exactly like `--no-docs` and only lists publications
without tags, fixes #237.

Currently publications without a tag have an empty set as the tags
variable. If this changes in the future (see #234) the conditional has
to be updated.


benureau:
Thanks! With each new functionality, we need even a small test to ensure coverage and verify that it does function appropriately across different versions and OS. We need to improve how easy it is to write test and the documentation for them. If you don't know how to do it, I will create a test for that PR.

PR #240, opened 2020-07-15
Add note on using commas in alias descriptions

StanczakDominik:
On my installation, using commas in docstring descriptions causes a `error: unsupported operand type(s) for %: 'list' and 'dict'` error. Wrapping the description in a string solves that.

### DIFF
--- a/pubs/config/spec.py
+++ b/pubs/config/spec.py
@@ -99,6 +99,11 @@ active = force_list(default=list('alias'))
 # command = !pubs list -k | wc -l
 # description = lists number of pubs in repo
 
+# To use commas in the description, wrap them in a "" string. For example:
+# [[[hellocount]]]
+# command = !pubs list -k | wc -l; echo "hello!"
+# description = "lists number of pubs in repo, greets the user afterward"
+
 [[git]]
 # The git plugin will commit changes to the repository in a git repository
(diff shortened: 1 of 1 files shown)


pubs/config/spec.py
```diff
@@ -99,6 +99,11 @@
 # command = !pubs list -k | wc -l
 # description = lists number of pubs in repo
 
+# To use commas in the description, wrap them in a "" string. For example:
+# [[[hellocount]]]
+# command = !pubs list -k | wc -l; echo "hello!"
```
omangin, line 104: Minor comment here: for the sake of conciseness, I would suggest keeping only the first and last line of your change (no need for a new fake section and a new fake comment).

omangin (changes requested):
Thanks @StanczakDominik for taking the time to edit the config documentation, and sorry that it took us so long to answer. See my minor comment below. Other than that, it is good to merge.

pubs/config/spec.py
StanczakDominik: Done :)

omangin (approved):
Great! Thanks @StanczakDominik!

omangin:
Note: you might have to rebase on master once #245 is merged for the tests to pass.

benureau:
Rebased and merged to main. Apologies for the delay.

PR #241, opened 2020-08-07
Allow passing named arguments to custom commands (alias).

jkulhanek:
This allows passing named arguments to custom commands. An example would be `pubs search keyword --ignore-author` and the corresponding alias would redirect the call to the custom script like follows: `search = !"$DOTFILES/pubs/scripts/search.py" "$@"`.

### DIFF
--- a/tests/test_plug_alias.py
+++ b/tests/test_plug_alias.py
@@ -1,7 +1,7 @@
 import shlex
 import unittest
-
 import dotdot
+import argparse
 
 import pubs
 from pubs import config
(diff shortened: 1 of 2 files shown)


omangin (commented):
Thanks @jkulhanek  for fixing this! and sorry for the late review.

Do you think you could add a quick unit test to https://github.com/pubs/pubs/blob/master/tests/test_plug_alias.py to prevent regressions?

jkulhanek:
A simple test added.

tests/test_plug_alias.py
```diff
@@ -61,6 +61,19 @@ def testShellAliasEscapes(self):
             shlex.split(self.subprocess.called.splitlines()[-1])[1:],
             args)
 
+    def testShellAliasNamedArguments(self):
+        parser = argparse.ArgumentParser()
+        parser.add_argument('--test2')
```
omangin, line 66: Did you intend to use this one?

omangin (approved):
See minor comments. Thanks @jkulhanek  for adding the test!

benureau:
Rebased an pushed to master as 2e939dbc and 0777a99d30. Apologies for the long delay.

PR #253, opened 2021-01-17
Fix collision of entry type key with valid field name 'type'.

### LINKED ISSUE #252: Support field name "type"
Biblatex has an entry type `thesis` where you "use the `type` field to specify the type of thesis."

Pubs removes that field when specified, leaving an invalid entry because `type` is a required field for `thesis`. I think
https://github.com/pubs/pubs/blob/89429a8c24d0aef444de614f2a102feeb14b615f/pubs/bibstruct.py#L10
should use `'ENTRYTYPE'` instead, just like bibtexparser. See https://github.com/sciunto-org/python-bibtexparser/pull/42 and https://github.com/sciunto-org/python-bibtexparser/pull/53.

omangin:
Fixes #252.

### DIFF
--- a/tests/test_endecoder.py
+++ b/tests/test_endecoder.py
@@ -185,14 +185,6 @@ class TestEnDecode(unittest.TestCase):
         with self.assertRaises(decoder.BibDecodingError):
             decoder.decode_bibdata("@misc{I am not a correct bibtex{{}")
 
-    def test_endecode_preserves_type(self):
-        decoder = endecoder.EnDecoder()
-        entry = decoder.decode_bibdata(bibtex_raw0)
-
-        bibraw1 = decoder.encode_bibdata(
-            entry, ignore_fields=['title', 'note', 'abstract', 'journal'])
-
-
 
 if __name__ == '__main__':
     unittest.main()
--- a/tests/test_endecoder.py
+++ b/tests/test_endecoder.py
@@ -57,6 +57,12 @@ class TestEnDecode(unittest.TestCase):
 
         self.assertEqual(bibraw1, bibraw2)
 
+    def test_decode_bibtex_preserves_type_field(self):
+        """Test that multiple encode/decode step preserve data"""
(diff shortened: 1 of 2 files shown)


tests/test_endecoder.py
```diff
@@ -179,6 +185,14 @@ def test_endecodes_raises_exception(self):
         with self.assertRaises(decoder.BibDecodingError):
             decoder.decode_bibdata("@misc{I am not a correct bibtex{{}")
 
+    def test_endecode_preserves_type(self):
+        decoder = endecoder.EnDecoder()
+        entry = decoder.decode_bibdata(bibtex_raw0)
+
+        bibraw1 = decoder.encode_bibdata(
+            entry, ignore_fields=['title', 'note', 'abstract', 'journal'])
+
```
benureau, line 194: wouldn't you want to re-decode bibraw1 at that point to make sure `type` is still present and "technical report"?

benureau (approved):
All tests pass on my machine. Beside the test that maybe could be extended, it's good.

tests/test_endecoder.py
omangin: Indeed, I think I left this one unfinished.

tests/test_endecoder.py
```diff
@@ -57,6 +57,12 @@ def test_endecode_bibtex(self):
 
         self.assertEqual(bibraw1, bibraw2)
 
+    def test_decode_bibtex_preserves_type_field(self):
+        """Test that multiple encode/decode step preserve data"""
+        decoder = endecoder.EnDecoder()
+        entry = decoder.decode_bibdata(bibtex_raw0)
+        self.assertEqual(entry['Page99']['type'], "technical report")
```
dotlambda, line 64: Should the result be `"Technical Report"`? It doesn't seem to preserve case.

dotlambda: See https://github.com/pubs/pubs/pull/262.

PR #260, opened 2021-01-26
Fix/256

### LINKED ISSUE #256: Fix continuous integration
Whether by fixing Travis or switching to Github Actions.

omangin:
Migrate from Travis to Github actions.

- [x] Create github action
- [x] Include install test and condition to being on schedule
- [x] Remove travis file
- [ ] Check that install test runs on schedule (can be done only after merge)
- [x] Add badge to wiki
- [x] Remove remaining travis config (webhook etc.)

Fixes #256.


tests/test_git.py
```diff
@@ -72,6 +83,7 @@ def test_git(self):
         # self.assertEqual(hash_i, hash_j)
 
     def test_manual(self):
+        print(self.default_pubs_dir)
```
benureau, line 86: Should that be there?

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - cron: '0 8 * * *'
+
+
+jobs:
+  unit-test:
+    name: Run unit tests
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest]
+        python-version: [3.6, 3.7, 3.8, 3.9]
```
benureau, line 16: We are not testing Python 2.7 anymore. Should be start updating the readme to remove mention of Python 2 support, and then actually remove it in a future release?

benureau (approved):

benureau:
Nicely done!

tests/test_git.py
omangin: Good catch! Thanks.

.github/workflows/tests.yaml
omangin: Actually the readme already states that it is not supported any more as we had already stopped testing for it some time ago. We could indeed remove all mentions of it with the next release. I will leave it for now anyway.

PR #262, opened 2021-02-27
preserve case of field 'type'

dotlambda:
Non-standard types such as "Diplomarbeit" should not be converted to lowercase. See e.g. https://github.com/jgm/pandoc-citeproc/issues/98.


benureau (approved):
Thank you for the review. Indeed, in German, that will create problems. I am merging this. I will keep it in mind if/when switching to a new parser.

PR #265, opened 2021-04-05
New features: show and search (and enhanced tag display)

richoux:
Hi there,

Pubs was the right tool for me, I was very happy to found it! However, I though 2 functionalities were missing:
1. just printing on the screen an entry (giving its citekey), 
2. searching in the repo for entries, giving an author name and/or some keywords from the title.

So I added them:
`$ pubs show MyCitekey`
will print on the screen the same pretty line of information than the one printed just after adding a new entry.

`$ pubs search --author ThisDude` (or -a ThisDude)
will display all papers where ThisDude is a (co-)author

`$ pubs search --title "Look for these keywords in title"` (or -t "Look for these keywords in title")
will display all papers containing the block "Look for these keywords in title" in the title. It will not look for each keyword separately, it is more like one keyword that can be a block of words with spaces. This can be changed of course, but so far it covers my needs.

`$ pubs search -a ThisDude -t "Look for these keywords in title"`
will display all papers where ThisDude is a (co-)author and containing "Look for these keywords in title" in the title.

Search is case-insensitive, on purpose.

I also changed how tags are displayed (both with `$ pubs tag` and with `$ pubs tag MyCitekey`), adding an uncolored comma between tags. I have tags composed of two words separated by one space, so just separating tags by a space was not convenient at all.
I have updated tests so it can work with this new way of printing tags.

### DIFF
--- a/pubs/commands/__init__.py
+++ b/pubs/commands/__init__.py
@@ -8,6 +8,8 @@ from . import remove_cmd
 from . import list_cmd
 from . import edit_cmd
 from . import tag_cmd
+from . import search_cmd
+from . import show_cmd
 from . import statistics_cmd
 # doc
 from . import doc_cmd
--- a/pubs/commands/import_cmd.py
+++ b/pubs/commands/import_cmd.py
@@ -21,7 +21,7 @@ _IGNORING_MSG = " Ignoring it."
 def parser(subparsers, conf):
     parser = subparsers.add_parser(
         'import',
-        help='import paper(s) to the repository.')
+        help='import paper(s) to the repository')
     parser.add_argument(
         'bibpath',
         help=("path to bibtex, bibtexml or bibyaml file, or a directory "
--- a/pubs/commands/search_cmd.py
+++ b/pubs/commands/search_cmd.py
@@ -1,0 +1,60 @@
+from __future__ import unicode_literals
+
+from ..paper import Paper
+from .. import repo
+from .. import color
+
+from ..uis import get_ui
+from ..endecoder import EnDecoder
+from ..utils import resolve_citekey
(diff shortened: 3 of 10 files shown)


benureau:
@richoux: Hello, sorry for the terrible delay in attending this PR. Thank you for the contribution.

Most of the functionality you implemented can be found in the `list` command. `pubs show MyCitekey` is equivalent to `pubs list citekey:MyCitekey`, and `pubs search --author ThisDude` to `pubs list author:ThisDude`, and these can be combined as well. That's really our fault, the documentation (i.e. the readme) does not mention this at all. We need to improve this. Could you tell me if the `pubs list` command fits all your needs? We won't duplicate functionality like that, but if there are improvement you made with `pubs seach` we could merge them to `pubs list`.

I was thinking about implementing (one day) a more "seach engine" like `pubs search` functionality, where we would just type a query, for instance `pubs search Bengio transformers` , and we would get a list of results, with hopefully https://arxiv.org/abs/2103.00336 near the top. 

The tag printing is indeed quite needed if you have tags with space. I will look into that. Tell me if you are still interested in having that merged into master.

richoux:
Hi @benureau, no problem for the delay, there are nothing to hurry about. :)

I wasn't aware of these possibilities with `pubs list` indeed! The command `pubs list --help` is fairly complete, but I think that the description with `pubs --help` should be more specific about `list` possibilities, in addition to a more detailed Readme.

No need to merge then! For the tag printing, you can just cherry-pick the code I wrote about that (copy/pasting will do the trick, no need that I appear as a contributor).

benureau:
The relevant commits have been cherry-picked and merged with master in #275. Closing this.

PR #278, opened 2022-08-10
make tests compatible with bibtexparser 1.3.0

dotlambda:
See https://github.com/sciunto-org/python-bibtexparser/commit/a8527ff.


benureau:
@dotlambda: thanks for the fixes. Merging this.

Issue #276, opened 2022-04-17
OttoBib is no longer in service, breaking ISBN lookup


meanderingexile:
Adding a new entry with the `-I {ISBN}` produces the standard error message:

```
error: ('404 Client Error: Not Found for url: https://www.ottobib.com/isbn/9780521705455/bibtex',)
```

Checking [https://www.ottobib.com/](https://www.ottobib.com/), it looks like the service has been depreciated. They suggest some alternative sources, such as [https://zbib.org/](https://zbib.org/). 

Would it be possible to update this service? It seems like it should just be a matter of updating the hardcoded string in `isbn2bibtex`.

For the time being I've just been adding/fetching bibtex data manually, but will try to submit a PR if no one gets around to this first. 

Thanks for maintaining pubs!

benureau:
@meanderingexile: Thanks for the info. I have looked at zbib.org, but not found any straigthforward way to get the bibtex from it from python (it is probably doable, not sure that (1) it would be robust (2) zoterobib would like that). If you are up to look into it and submit a PR, that would be very useful.

benureau:
In the meantime, trying to add an ISBN will offer a error with an explanation about the situation.

meanderingexile:
Just as a quick update on this: 

I now see what you mean about zbib! 

I ended up getting a local patch working using Internet Archive's OpenLibrary project as a data source. They offer a free low-volume API (documented here: [https://openlibrary.org/developers](https://openlibrary.org/developers)) and I believe any reasonable `pubs` user would be within their TOS.  They do not provide BibTeX directly, but munging the JSON was fairly straightforward.

If I don't discover any problems with this approach through my own use in the next few days I will submit a PR.

chrberrig:
both open library, as well as worldcat seems do do the trick in regard to the databases:

https://www.vinzius.com/post/free-and-paid-api-isbn/

some additional parsing of the API-retuen val. might be neccessary though, but i guess that could be dome locally in pubs?
The WorldCat DB seems to be the bigger of the two.

benureau:
@meanderingexile, @chrberrig: thank you. I think that indeed OpenLibrary and WorldCat are both options that may be viable.

chrberrig:
made an attempt at fix, look at #280

PR #280, opened 2023-04-06
attempted isbn fix with the wikipedia REST-api, Issue #276

chrberrig:
So, this is a first attempt at fixing the ISBN fetching problem from back when ottobib shut down. 

as it stands, it seem the testing environments, still throws some errors at me, namely the following:

- for the case of isbn 9999999999999, which should result in a "no book found" err, the Wikipedia REST-api still retreives bibtex info on something. 

- for the Poincaré case (isbn 9782081336742), it seems that all the accents has moved along 1 letter, could this be a UTF-8 thing?...

for any of the above err.s that makes these tests not pass, i'm not sure why. But there are probably some of you guys smarter than I that can figure it out, and if I can be of further help, let me know :)


chrberrig:
so i'm stuck at the following error using pytest: 
TypeError: decode() got an unexpected keyword argument 'encoding'

this seem to refer back to the bibtexparser, wich i have no experience with, along the encoding module made for pubs, as i read it. 

any ideas?

chrberrig (commented):
Some of the comments are obviously from the ottobib days (l. 100, 106, 107, 109), and should be removed. Also the test for nonsense isbn should be reintroduced at some point, but for valid isbn, the isbn bibtex fetching should be working now!

chrberrig:
obvious cleanup is now done...

benureau:
Thank you for the PR for this important feature. I will review and test it over the next few days.

benureau:
Based on my experimentation, it seems the citation API from Wikipedia is doing a lot of fuzzy research across multiple fields and reference numbers (DOI, ISBN, PUBMED...). For instance, asking for (presumably) ISBN 9999999999999999111 returns a result for ISBN 9999999999:
`$curl -X 'GET' 'https://en.wikipedia.org/api/rest_v1/data/citation/mediawiki/9999999999999999111' -H 'accept: application/json; charset=utf-8;'`
`[{"itemType":"book","title":"Nithān phư̄a kānrīanrū phāsā Lānnā : Tao nō̜i ʻō̜ngkham","oclc":"893209173","url":"https://www.worldcat.org/oclc/893209173","edition":"Phim khrang thī 1","place":"Chīang Mai","date":"2557 [2014]","numPages":"1 volume (unpaged)","contributor":[["","Mahāwitthayālai Chīang Mai. Samnak Songsœ̄m Sinlapa Watthanatham"]],"author":[["","Natthaphong Pančhaburī"],["","ณัฐพงษ์ ปัญจบุรี"]],"accessDate":"2023-04-16","ISBN":["9999999999"],"source":["WorldCat"]}]`

It's actually difficult to get a 404 response from the service. `0` returns a match based on the title. 
`curl -X 'GET'   'https://en.wikipedia.org/api/rest_v1/data/citation/mediawiki/0'   -H 'accept: application/json; charset=utf-8;'`
`[{"itemType":"audioRecording","title":"@0","url":"http://worldcat.org/oclc/1269444707","oclc":"1269444707","author":[["Coldcut (Musical group), remix artist,","performer."]],"accessDate":"2023-04-16","source":["WorldCat"]}]`

After trying different things that all returned matches, I generated a random set of letters and digits `Hah4aewu`, which returned a 500 status error, not 404.
`curl -X 'GET'   'https://en.wikipedia.org/api/rest_v1/data/citation/mediawiki/Hah4aewu'   -H 'accept: application/json; charset=utf-8;'`
`{"type":"https://mediawiki.org/wiki/HyperSwitch/errors/unknown_error","method":"get","uri":"/en.wikipedia.org/v1/data/citation/mediawiki/Hah4aewu"}`

I think the service is quite useful, and since ISBN API are rare, we can't be choosy. But we should be handling the response a bit more. First, we need to check for status 200 on the response. Anything else should be an error. Then we should check that the bibtex is for the ISBN we asked for. Anything else should be an error too, since the matching is way to permissive (`0` returns a match, `query` returns a match). What do you think @chrberrig?

PR #283, opened 2024-12-23
Replaced pipes with shlex

### LINKED ISSUE #282: Python 3.13 removed 'pipes' module
With Python 3.13, pubs fails to run with the message `error: No module named 'pipes'`. This is because the pipes module was removed in 3.13, see e.g. [this issue for jupyter](https://github.com/jupyter/nbclassic/issues/308).

kentdobias:
Python 3.0 deprecated the pipes module and Python 3.13 removed it. The shlex module provides a drop-in replacement for the quote function from pipes that is used by pubs. This resolves [https://github.com/pubs/pubs/issues/282](https://github.com/pubs/pubs/issues/282).

### DIFF
--- a/pubs/plugs/alias/alias.py
+++ b/pubs/plugs/alias/alias.py
@@ -1,7 +1,7 @@
 import shlex
 import subprocess
 import argparse
-from pipes import quote as shell_quote
+from shlex import quote as shell_quote
 
 from ...plugins import PapersPlugin
 from ...pubs_cmd import execute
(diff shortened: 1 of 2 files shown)


bobismijnnaam:
I think this is pretty safe to merge. This is a pretty safe workaround: even in 3.12 the pipes module itself [used the version from shlex](https://github.com/python/cpython/blob/8cb2092bd8cfe678829ff4b665f39bfa7bd3b6ac/Lib/pipes.py#L66). If it can't be merged, can the maintainers comment on what else needs to happen?

PR #285, opened 2026-05-04
fix: replace deprecated pipes.quote with shlex.quote (Python 3.13 compat)

### LINKED ISSUE #282: Python 3.13 removed 'pipes' module

wali-reheman:
## Summary

The `pipes` module was removed in Python 3.13 (PEP 594). This PR replaces `from pipes import quote as shell_quote` with `from shlex import quote as shell_quote` in the two files that use it:

- `pubs/plugs/alias/alias.py`
- `pubs/plugs/git/git.py`

`shlex.quote` has been in the stdlib since Python 3.3 and is the direct replacement with identical shell-escaping behavior.

## Test Plan

- ✅ Verified both files import cleanly with `python -c "import ..."` (requires deps not present in this env, but the change is a pure stdlib swap — no behavioral change)
- ✅ Manual diff review confirms only the import line changed in each file

Closes #282

### DIFF
--- a/pubs/plugs/alias/alias.py
+++ b/pubs/plugs/alias/alias.py
@@ -1,7 +1,7 @@
 import shlex
 import subprocess
 import argparse
-from pipes import quote as shell_quote
+from shlex import quote as shell_quote
 
 from ...plugins import PapersPlugin
 from ...pubs_cmd import execute
--- a/pubs/plugs/git/git.py
+++ b/pubs/plugs/git/git.py
@@ -2,7 +2,7 @@ import os
 import sys
 import argparse
 from subprocess import Popen, PIPE, STDOUT
-from pipes import quote as shell_quote
+from shlex import quote as shell_quote
 
 from ... import uis
 from ...plugins import PapersPlugin


pubs/plugs/alias/alias.py
```diff
@@ -1,7 +1,7 @@
 import shlex
 import subprocess
 import argparse
-from pipes import quote as shell_quote
+from shlex import quote as shell_quote
```
chrberrig, line 4: Hi there, 

so for backward compat, may i suggest
```
try:
    from pipes import quote as shell_quote
except ImportError:
    from shlex import quote as shell_quote
```

I actually did this in PR #286, but did not see your issue. 

My bad. 

But i think this would make the transition safer

pubs/plugs/git/git.py
```diff
@@ -2,7 +2,7 @@
 import sys
 import argparse
 from subprocess import Popen, PIPE, STDOUT
-from pipes import quote as shell_quote
+from shlex import quote as shell_quote
```
chrberrig, line 5: Similar as above

